PR #48, opened 2017-06-02
WIP: Challenge listing refactoring

birdofpreyru:
- Initial loading of challenges into the listing is done with Redux
actions / reducers / state;

- My challenges are fetched again.

- If navigation around the listing page triggers loading of additional
challenges, it is not handled properly (old fetch code is executed,
which can be kind of fine for the stand-alone listing, but definitely
not good for the listing inside TC communities).

- Challenge counts in the sidebar are wrong.

- In general there is a lot of room for further improvements.

### DIFF
--- a/src/shared/actions/challenge-listing.js
+++ b/src/shared/actions/challenge-listing.js
@@ -1,13 +1,133 @@
 /**
  * Challenge listing actions.
+ *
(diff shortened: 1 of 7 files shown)


ajefts:
nice work @birdofpreyru 

Once thing I notice is if you filter on a keyword, and then remove that keyword it loads all the public challenges again..

Issue #170, opened 2017-08-23
[$120] - Challenge Details - Checkpoints - Refactor the code


birdofpreyru:
The code in `/src/shared/components/challenge-detail/Checkpoints/index.jsx` does work, but is hacky, thus should be refactored. The fix should address the following:

- Right now the component uses local state, window resize event, and some hacky code to "trim" the content when page size goes under 768 px. The same should be achieved by pure SCSS (using media mixins defined in `/src/styles/mixins/_media.scss`.

- Also collapse / expansion of the checkpoint sections is handled using the local state. It should be refactored to properly use Redux store / actions / reducers.

daft300punk:
Please assign me this.

birdofpreyru:
@daft300punk Consider yourself assigned. I have just noticed that I don't have enough rights in the repo to add contributors, and without it, I cannot technically assign people who are not contributors. Once my rights corrected, I'll fix it.

daft300punk:
Check out the PR - https://github.com/topcoder-platform/community-app/pull/229

daft300punk:
Is it a bug? The original site doesn't do anything when clicked on those side ids. Anyways, it is easy to fix. I will submit PR as soon as you confirm.

daft300punk:
Updated the PR, take a look. https://github.com/topcoder-platform/community-app/pull/229

daft300punk:
Updated the PR, take a look https://github.com/topcoder-platform/community-app/pull/229

1. Removed jquery dependency
2. To check scroll to feedback, click on the first 4 feedback ids in the side bar. This will increase page size. Now, go to top, and then click on the 5th feedback id. Should scroll down to it.

PR #229, opened 2017-08-24
Challenge Details - Checkpoints - Refactor the code #170

daft300punk:
1. Click on feedback header. Check dispatched action TOGGLE_CHECKPOINT_FEEDBACK. No internal state, all managed by redux.
2. Removed trim logic, just pure scss.


birdofpreyru:
l@daft300punk Generally Looks good, though I want the following small corrections:
- `decodeEscaped(..)` function is really goofy and unnecessary. I believe, we can just use `dangerouslySetInnerHtml` prop instead?
- It looks also that nothing else depends directly on `jquery` in our app? So, `jquery` dependency should be removed.
- In desktop view, when clicking on submission buttons at left, let's in addition to toggling the section, also scroll to it (it also makes sense to make those button just open and scroll, rather than toggle the section. i.e. if the section is already opened, clicking that button just scrolls to it).

src/shared/components/challenge-detail/Checkpoints/index.jsx
```diff
@@ -1,19 +1,15 @@
 import React from 'react';
 import PT from 'prop-types';
 
-import jquery from 'jquery';
```
birdofpreyru: Ok, but what about removing `jquery` from `package.json`, `package-lock.json` and `.eslintrc.json`? I believe, it is not used anywhere else in our app, and it is not supposed to be. Do you see anything wrong with it?

src/shared/components/challenge-detail/Checkpoints/index.jsx
```diff
@@ -1,19 +1,15 @@
...
-  return jquery('<textarea />').html(escaped).text();
-}
-
 const Checkpoints = (props) => {
   const {
     checkpointResults,
     generalFeedback,
   } = props.checkpoints;
+
+  /* eslint-disable react/no-danger, no-undef */
```
birdofpreyru: 1. Please, avoid disabling eslint rules for entire file. Either disable them for specific lines, adding `// eslint-disable-line react/no-danger` to the end of that line; or, if this results in a way too long line (> 80 columns), disable it right before the line(s) where it is used, and re-enable right after.

2. Don't disable `no-undef` rule, just add to the beginning of the file (under header comment, if any) `/* global document */`

src/shared/components/challenge-detail/Checkpoints/index.jsx
daft300punk: It is in fact used somewhere else, i.e. not directly but by a package react-slick. I'm removing it from package.json, but it will remain in package-lock.json.

birdofpreyru: Ha! `package-lock.json` mentions separately package versions installed as direct dependencies, and those used as dependencies of dependencies, so it is actually straightforward to just remove what we want to remove. Though, don't worry, I'll do this last small fix myself, and if everything works fine, I'll merge the PR.

Issue #132, opened 2017-07-31
[$20] - Challenge Listings: Minor styling bugs at the XXS (<320px) sreen size


birdofpreyru:
![screenshot](https://user-images.githubusercontent.com/20144632/28779986-70106fd8-7605-11e7-96c2-7a8339ecea9b.png)

![screenshot](https://user-images.githubusercontent.com/20144632/28780019-94eb974c-7605-11e7-9952-f8c2ec6e9e39.png)

elfman:
can you assign me this?

birdofpreyru:
@elfman Consider yourself assigned. I have just noticed that I don't have enough rights in the repo to add contributors, and without it, I cannot technically assign people who are not contributors. Once my rights corrected, I'll fix it.

elfman:
@birdofpreyru what is the min screen size this page should compatible with?

birdofpreyru:
@elfman 0px :-D
Well, we are fine if staff just does not fit into the screen cause there is not enough space. But, in the screenshots above, my problem is that `Tracks` and `Filters` loose background and hover over other elements, rather than just occupy additional row. And in the filters panel, everythings just is not aligned properly to occupy available width.

elfman:
I have raised a pr, but I can't change the issue's label

birdofpreyru:
@elfman I've send you invitation to the repo. I guess, you should check your email, accept it, and then you'll be able to update the labels. I'll check your PR later :)

PR #246, opened 2017-08-25
Fix issue #132

elfman:


birdofpreyru:
@elfman I am not 100% sure, whether this PR is your final solution to the issue, but I see here two small things to correct further:
screenshot
`Sub community` and `Subtrack`, `Date range` labels need a bit less left padding, to be on the same vertical line (true for larger screen size as well - all elements should be aligned at the same vertical line).
`Clear filters` button, needs a bit more left padding to be on the same line.

Also, on very small screen:
screenshot

elfman:
Updated

Issue #200, opened 2017-08-24
[$60] - Underscore displayed in the 'First 2_finish' and 'Ui Prototype_competition' tags


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges/30058991/?type=develop

- Check the tag > First 2_finish

- Go to http://community-app.topcoder.com/challenges/30058932/?type=develop

- Check the tag > Ui Prototype_competition

Actual Result(s)
-----------------------------------------------------------------
- First 2_finish
- Ui Prototype_competition

Expected Result(s)
-----------------------------------------------------------------
- First 2 finish
- UI Prototype

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![13](https://user-images.githubusercontent.com/4592712/29670844-95c03b1a-8905-11e7-8e79-a3d2e99a1518.jpg)

birdofpreyru:
Accepted, 3 points.
### Expected Fix
I believe, since the original implementation of the page, the backend has been updated to return consistent names of subtracks, it should be double-checked, and updated in the frontend accordingly.

chqy24:
Assign to myself
and please grant me permission to change the label, thanks

chqy24:
Ready to review, PR #247

chqy24:
@birdofpreyru 
PR #247 has been updated

PR #247, opened 2017-08-25
fix issue #200

chqy24:
Fix issues #200

### DIFF
--- a/__tests__/shared/actions/challenge-listing/index.js
+++ b/__tests__/shared/actions/challenge-listing/index.js
@@ -91,7 +91,7 @@ describe('challengeListing.getChallengeSubtracksDone', () => {
   });
 
   test('payload is a promise which resolves to the expected object', () =>
-    a.payload.then(res => expect(res).toEqual([{"description": "dummy"}])));
+    a.payload.then(res => expect(res).toEqual([{ description: 'dummy' }])));
 });
 
 describe('challengeListing.getChallengeTagsDone', () => {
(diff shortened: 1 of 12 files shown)


birdofpreyru:
@chqy24 Nope, here we need a better fix. Look, the api v3 has been updated to get the following info for possible challenge types:
screenshot

Now, there is action / reducer to get challenge types for filters in challenge listing (maybe somewhere else), which uses v2 api. We should update those to use v3 api instead, without breaking existing code. Once we have it, we can get `subTrack` from v3 challenge details as it is, look for the matching challenge type, and get it proper human-readable name to render in the tag. Ideally, to make it efficient, when we fetch challenge types we should store them not only as an array (for challenge listing kind of use), but also like a map between subTrack and the object, so that we can lookup it efficiently.

birdofpreyru:
@chqy24 Now it looks mostly cool, and challenge details page is perfect.
Though, I see some bugs with challenge details page:
1. When I select subtracks in the subtrack filter's dropdown for many of them no challenges match;
2. When I click subtrack tag in a challenge details page, it does take me to the challenge listing, with proper filter applied, and the filters count is set correctly, however, no filter is selected in the subtrack filter's dropdown (though filtering works properly even for those subtracks, that do not filter when I do 1).
Please, look into it to make it work properly. I expect, the fix should be trivial.

chqy24:
I'd noticed these issues, but these are existed bugs and should be separated from #200

birdofpreyru:
@chqy24 I really want them to be fixes as a part of this PR, so that we don't loose those a track of these issues. Under the hood, I believe, the only problem is that while now you properly fetch and use v3 subtrack keys in the challenge details page, in the challenge listing's dropdown the values from v2 are still used, so it causes the problem, as challenge objects, most probably, have sub track keys from v3.

So, the fix should be very easy. I can add $20 to the payment for cleaning this out.

chqy24:
Sorry, you're right, The first issue was introduced by my previous commit
Please check it now.

Issue #207, opened 2017-08-24
[$40] - register button  enabled for winner annonced challenge 


karthikbecse:
Steps to Reproduce

   
  Goto http://community-app.topcoder.com/challenges/30058908/?type=develop
  Click  winner link 
  Observe  Next Deadline:  and register button 
   

Actual Result(s)
Next Deadline:  staus is register and register button  enabled 

Expected Result(s)
Should show the status properly 

  
Environment

    Device(s): Laptop
    Operating System: Windows 8.1  64bit
    Browser(s): FF 54.0.1

Image/Video/JS Log/Console Log 
![screenshot-community-app topcoder com 2017-08-24 20-03-53](https://user-images.githubusercontent.com/3225433/29671983-a75db2aa-8908-11e7-9dc4-c034b342b370.png)

chqy24:
ready for review, PR #266

chqy24:
@birdofpreyru 
PR #266 has been updated

PR #266, opened 2017-08-25
fix issue #207

chqy24:
issue #207


birdofpreyru:
@chqy24 Please update to ensure that in such case we also properly write that the challenge is completed in the `Next Deadline` section, and in the collapsible section with deadline details we also show the correct end date, if we can get it, or at least we mark the end date with green color, as completed.

Issue #212, opened 2017-08-24
[$30] - [Bug Hunt - NEW BUG]Can't handle invalid date properly


twicoder:
1) Open URL
http://community-app.topcoder.com/challenges/?filter[text]=&filter[endDate]=99999999999-09-01T04:00:00.000Z
2) The date is invalid, but it should be handled properly

birdofpreyru:
Accepted, 3 points.
Indeed, we should validate that dates suggested in URL query are proper, otherwise, ignore them and remove corresponding element of the query.

twicoder:
ready for review, PR #272

PR #272, opened 2017-08-25
Fix issue #212

twicoder:
The fix will check the startDate and endDate, if date is invalid, then ignore them in the search logic.

### DIFF
--- a/src/shared/components/challenge-listing/index.jsx
+++ b/src/shared/components/challenge-listing/index.jsx
@@ -16,14 +16,12 @@ import Listing from './Listing';
 import ChallengeCardPlaceholder from './placeholders/ChallengeCard';
 import SRMCard from './SRMCard';
 
-import { updateQuery } from 'utils/url';
 import './style.scss';
 
 // Number of challenge placeholder card to display
 const CHALLENGE_PLACEHOLDER_COUNT = 8;
 
 export default function ChallengeListing(props) {
-  updateQuery({ });
   let challenges = props.challenges;
 
   if (props.communityFilter) {
--- a/src/shared/components/challenge-listing/index.jsx
+++ b/src/shared/components/challenge-listing/index.jsx
@@ -22,6 +22,29 @@ import './style.scss';
 const CHALLENGE_PLACEHOLDER_COUNT = 8;
 
 export default function ChallengeListing(props) {
+  function CheckDateTime(str) {
+    const reg = /^(\d{4})-(\d{1,2})-(\d{1,2})T(\d{1,2}):(\d{1,2}):(\d{1,2})\.(\d{3})Z$/;
+    const r = str.match(reg);
+    if (r == null) return false;
+    r[2] -= 1;
+    const d = new Date(r[1], r[2], r[3], r[4], r[5], r[6]);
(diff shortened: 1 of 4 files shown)


src/shared/components/challenge-listing/index.jsx
```diff
@@ -22,6 +22,29 @@ import './style.scss';
...
+    const d = new Date(r[1], r[2], r[3], r[4], r[5], r[6]);
+    if (d.getFullYear() !== r[1]) return false;
+    if (d.getMonth() !== r[2]) return false;
+    if (d.getDate() !== r[3]) return false;
+    if (d.getHours() !== r[4]) return false;
+    if (d.getMinutes() !== r[5]) return false;
+    if (d.getSeconds() !== r[6]) return false;
+    return true;
+  }
+  const filterState = props.filterState;
```
birdofpreyru: @twicoder And, although it probably will do the trick, the best way to make the check and avoid using the date in filter, if it is invalid, is inside reducer (the filter query params are picked up from URL during server-side rendering, thus checking it there will guarantee that invalid dates won't be used anywhere): https://github.com/topcoder-platform/community-app/blob/develop/src/shared/reducers/challenge-listing/index.js#L272

Also, once at the client-side, we should remove invalid dates from query. This can be done at this point. To update query params use this utility function: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/utils/url.js#L20

twicoder:
@birdofpreyru  I have changed the code as suggested.

src/shared/reducers/challenge-listing/index.js
```diff
@@ -270,6 +271,14 @@ export function factory(req) {
 
   if (req) {
     state.filter = req.query.filter;
+    if (!!state.filter && !!state.filter.startDate
+        && moment(state.filter.startDate).isValid() === false) {
```
birdofpreyru: Hmm... It is enough to write
`if (state.filter && state.filter.stateDate && moment(state.filter.startDate).isValid()) {`
not sure, why you don't do it this way? Please fix.

src/shared/utils/url.js
```diff
@@ -28,6 +29,13 @@ export function updateQuery(update) {
     if (_.isUndefined(value)) delete query[key];
     else query[key] = value;
   });
+  if (!!query.filter && !!query.filter.startDate
+      && moment(query.filter.startDate).isValid() === false) {
+    delete query.filter.startDate;
+  } else if (!!query.filter && !!query.filter.endDate
+      && moment(query.filter.endDate).isValid() === false) {
+    delete query.filter.endDate;
+  }
```
birdofpreyru: No, you was not supposed to modify `updateQuery(..)`
what I was talking about was to make such checks in `src/shared/routes/Topcoder/ChallengeListing.jsx`
and if you see we need to remove these query params, then you do it by calling `updateQuery({ startDate: undefined, endDate: undefined })`. The `updateQuery(..)` method is just an auxiliary function for setting / updating / deleting query params from URL. That's it.

twicoder:
@birdofpreyru  Changed as suggestion, thanks!

Issue #210, opened 2017-08-24
[$20] - [Common] Page won't focus on top once the user clicked the 'view more challenges' link in each sections


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges

- Click 'view more challenges' link in 'Open for registration'

- Click 'Ongoing challenges' link in 'Open for registration'

- Click 'Open for review' link in 'Open for registration'

- Observe the page position

Actual Result(s)
-----------------------------------------------------------------
- [Common] Page won't focus on top once the user clicked the 'view more challenges' link in each sections

Expected Result(s)
-----------------------------------------------------------------
- Must focus the page on top when user click on the 'view more challenges' link

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![17](https://user-images.githubusercontent.com/4592712/29672488-02a5230e-890a-11e7-87a0-9ce5aa9ebe58.jpg)

birdofpreyru:
Accepted, 3 points.
Indeed, when the buckets are switched in the sidebar, we should scroll to the top of the page.

daft300punk:
Check PR https://github.com/topcoder-platform/community-app/pull/273

daft300punk:
Updated PR - https://github.com/topcoder-platform/community-app/pull/273

PR #273, opened 2017-08-25
Issue #210

daft300punk:


src/shared/components/challenge-listing/index.jsx
```diff
@@ -85,6 +85,12 @@ export default function ChallengeListing(props) {
     );
   }
 
+  /* eslint-env browser */
+  if (document) {
```
birdofpreyru: @daft300punk To check, whether the code is executed at server or client side, please always use `isClientSide()` or `isServerSide()` from https://github.com/topcoder-platform/community-app/blob/develop/src/shared/utils/isomorphy.js

Also, placing this code into render method, I expect, will have side effects: ReactJS is free to re-render a component in many cases, not only when we enter the page, thus this code, I believe, will scroll the page to top in some cases when we don't need it. I guess, we should put it into `onClick` handler of the `View more challenges` button. Also, inside that handler you don't have to check for client side, as click handlers are always invoked at client-side only.

birdofpreyru:
@daft300punk any update to this?

birdofpreyru:
@daft300punk Hey-ho? Any update?!!

daft300punk:
Checkout now. Sorry for the delay, was a little away for visa purposes.

Issue #190, opened 2017-08-24
[$30] - [Bug Hunt - NEW BUG]Unexpected URL is not handled properly


twicoder:
1) try to open url: http://community-app.topcoder.com/challenges/*
2) Page body is not shown
<img width="1639" alt="2017-08-24 9 59 01" src="https://user-images.githubusercontent.com/3046795/29669897-87dbeb54-8917-11e7-8332-38b8e5c81433.png">

birdofpreyru:
Accepted, 3 points.

### Expected Fix
The routing should be corrected, to ensure that when a non-digital challenge ID is supplied in such route, it renders 404 Error page.

birdofpreyru:
Please also take into account #192 which is also caused by incorrect challenge ID used in URL. That should also render 404 error page.

elfman:
assign myself

elfman:
PR arised.
#192 is not fixed in this PR, server should return something when chanlengeId is not exist, currently it keeps the link waiting and nothing response.

birdofpreyru:
@ajefts @skyhit See the last comment. It looks like when we pass a non-existing challenge ID into the backend, it just stucks, rather than returning and error (404?) right away.

PR #274, opened 2017-08-25
Fix issue #190

elfman:


src/shared/containers/challenge-detail/index.jsx
```diff
@@ -56,6 +57,10 @@ class ChallengeDetailPageContainer extends React.Component {
     const { challenge, loadChallengeDetails, loadTerms,
       openTermsModal, authTokens, challengeId } = this.props;
 
+    if (isNaN(challengeId)) {
```
birdofpreyru: Please, replace all occurances of `isNaN(..)` by `_.isNumber(..)` from `lodash`; because https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/isNaN (especially, check `Confusing special-case behavior` section there)

elfman: the result of `_.isNumber(NaN)` is `true`, I will use `_.isNaN` instead

elfman:
I found some bug, I'm fixing it.

Issue #205, opened 2017-08-24
[$20] - [Bug Hunt - NEW BUG]Blank filters are shown in edit page but can't see in normal page


twicoder:
1) Create several filters and then edit them, rename their names to blank, and click "done"
2) You will not see the blank name filters but if you edit them, they will back
3) Maybe should forbid blank filter name
<img width="639" alt="2017-08-24 10 31 13" src="https://user-images.githubusercontent.com/3046795/29671519-58e5511e-891c-11e7-8abd-3affe79e5e82.png">

birdofpreyru:
Accepted, 3 points.
Indeed, we should not allow to save blank filter names.

twicoder:
PR #283

PR #283, opened 2017-08-25
Fix issue #205

twicoder:
If filter name is blank, it will not be updated.


birdofpreyru:
@twicoder 
1. Your change fails `npm test`.
2. I believe, it should be done a bit differently: if user tries to save fitlers when some of them has empty name, the `Done` button should be disabled, and the frame around empty filter name should become red (and a red text message `Cannot save filters with empty names` should be rendered somewhere near `Done` button.

birdofpreyru:
@twicoder any update here?

PR #293, opened 2017-08-25
fixed #222 corrected urls for tags

### LINKED ISSUE #222: [$20] - Clicking on tags will go to All challenges page
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges/30058990/?type=develop

- Click C++

Actual Result(s)
-----------------------------------------------------------------
- Clicking on tags will go to All challenges page

Expected Result(s)
-----------------------------------------------------------------
- Tags must not clickable or
- Must filter the challenges that have the clicked tags associated with (eg: If we click the tag Node.Js, must filter the challenges that have node.js associted with)

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

shubhendusaurabh:
fixed #222 corrected urls for tags

### DIFF
--- a/src/shared/components/challenge-detail/Header/ChallengeTags.jsx
+++ b/src/shared/components/challenge-detail/Header/ChallengeTags.jsx
@@ -28,7 +28,7 @@ export default function ChallengeTags(props) {
         subTrack &&
         <Link
           onClick={() => setChallengeListingFilter({ subtracks: [subTrack] })}
-          to="/challenges"
+          to={`/challenges?filter[subtracks][0]=${subTrack}`}
           styleName={`tag-common ${subTrackStyle}`}
         >{subTrack}</Link>
       }
@@ -48,7 +48,7 @@ export default function ChallengeTags(props) {
               <Link
                 key={tag}
                 onClick={() => setChallengeListingFilter({ tags: [tag] })}
-                to="/challenges"
+                to={`/challenges?filter[tags][0]=${tag}`}
                 styleName="tag-common misc-tag"
               >{tag}</Link>
             ),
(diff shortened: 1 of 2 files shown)


birdofpreyru:
@shubhendusaurabh Does not work for me :( It is a link inside app, handled by react-router without reloading the app (and this we don't wanna change), so just adding query params there has no effect.
Though, updating query params in URL is simple: we have this auxiliary function to do it: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/utils/url.js#L20
The only thing I am not sure, whether it will work if we add it to onClick handler, or we should add it to challenge listing component, to check and update query params when necessary on the component mount time.

birdofpreyru:
@shubhendusaurabh Any update here?

shubhendusaurabh:
@birdofpreyru updated as requested

birdofpreyru:
I'll pay you for this ticket, but I'll close this PR and merge in a different fix for this (b75f3b23202782136bddc79f35abfeaf7fa04b93).

In case you wonder why. I believe, the best place, in general, to sync query params in URL with Redux state are reducers (because it is also a good place to create appropriate intial Redux state for given query params in URL during server-side rendering). So, actually what we were wanting to do here was already handled here: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/reducers/challenge-listing/index.js#L170 the problem was, that that line set the correct query params, but the following transition to challenge listing with react-router's Link cleaned them out. So, the trick was just to do it in different order: first to trigger the transition, then to set necessary filter (and this automatically sets the necessary query params in reducer). Hence, `setImmediate` wrapper inside onClick.

Issue #220, opened 2017-08-24
[$20] - Challenge details page hyperlink (Hover, normal, visited) colors are wrong


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges/30058990/?type=develop

- Check the hyperlinks

Actual Result(s)
-----------------------------------------------------------------
- Challenge details page hyperlink (Hover, normal, visited) colors are wrong

Expected Result(s)
-----------------------------------------------------------------
- Must match with the style guide, attached

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![23](https://user-images.githubusercontent.com/4592712/29673616-6873273c-890d-11e7-8aee-307ae0eb10fa.jpg)

![23 1](https://user-images.githubusercontent.com/4592712/29673634-756e2e14-890d-11e7-8028-5bb7eb893e70.jpg)

birdofpreyru:
Accepted, 1 point

iversonLv:
For this page, spec content is written from Editor which has inline css, seems can't not fix via css file?

birdofpreyru:
Well, we don't touch those styles that have been explicitely set as inline styles by the editor. However, on this page it is not always the case. E.g. the first blue link becomes dark gray on hover, however I don't see any inline styling for this in the page source.

twicoder:
PR #301

PR #301, opened 2017-08-25
Fix issue #220

twicoder:
Add some styles for <a>

### DIFF
--- a/src/styles/_reset.scss
+++ b/src/styles/_reset.scss
@@ -170,20 +170,3 @@ a {
   }
 }
 
-article > a {
-  color: #3078df;
-}
-
-article > a:hover {
-  color: #3078df;
-  text-decoration: underline;
-}
-
-article > a:active {
-  color: #69a8f8;
-  text-decoration: underline;
-}
-
-article > a:visited {
-  color: #235496;
-}
--- a/src/styles/_reset.scss
+++ b/src/styles/_reset.scss
@@ -169,3 +169,21 @@ a {
     color: inherit;
   }
 }
+
+article > a {
+  color: #3078df;
+}
+
+article > a:hover {
+  color: #3078df;
+  text-decoration: underline;
+}
+
+article > a:active {
+  color: #69a8f8;
+  text-decoration: underline;
+}
+
+article > a:visited {
+  color: #235496;
+}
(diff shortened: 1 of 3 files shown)


src/styles/_reset.scss
```diff
@@ -169,3 +169,21 @@ a {
     color: inherit;
   }
 }
+
+article > a {
+  color: #3078df;
```
birdofpreyru: 1. The colors you use here do not really much design specs (https://drive.google.com/open?id=0B3y3SEA9cIEkRWt2TVpXMDhQbTA)!
2. Also, you should not put color codes directly to your custom styling. We have all proper colors defined as variables here: https://github.com/topcoder-platform/community-app/blob/develop/src/styles/_variables.scss#L43
So, you just check what is the variable name for that color you need, and put you scss `color: $tc-black;` or whatever you need.

twicoder:
@birdofpreyru  changed "$tc-dark-blue: #0a71e6;" to "$tc-dark-blue: #0b71e6;"

birdofpreyru:
@twicoder I guess you have misunderstood me. You sure don't touch the values of variables in _variables.scss - they are the standard colors from Topcoder styleguide, being used all around, they should stay the same.

What I ment, is that when you write a custom styling for hrefs rather than doing
```
article > a:hover {
  color: #3078df;
  text-decoration: underline;
}
```
you use those variables, like (provided that you have `@import "~styles/tc-styles";` on top of your `.scss`):
```
article > a:hover {
  color: $tc-dark-blue;
  text-decoration: underline;
}
```

Also I've just noticed that you attempted to put your custom styles into _reset.scss - this is also wrong. Inside `/src/styles` we have only the global styles and mixins. The actual styles for individual components are located along with the components code. In this very case you should modify this stylesheet, I believe: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/components/challenge-detail/Specification/styles.scss
to ensure that links in the challenge specifications (if they don't have inline styling) match the design.

twicoder:
@birdofpreyru  Changed scss file as suggested. Value #0b71e6 is from design specs.

birdofpreyru:
@twicoder  Btw, have you seen my last post in the challenge forum? I don't know what is your TC nickname to be able to pay for the Bug Hunt part of the challenge.

twicoder:
@birdofpreyru  My TC nickname is renqingwei123, I have post it in the forum. Thanks!

birdofpreyru:
@twicoder 
1. OK, this is correct use of standard colors. For regular link, though, you should just use the value of `$tc-dark-blue`: it is a bit different from design specs, but, I guess, it is just a misprint in specs.
2. The problem is that your css modification does not really help: when I go to the URL of challenge in the issue ticket, scroll to the first blue link, and hover, it becomes black (very dark blue), it is different from specs.

Issue #175, opened 2017-08-24
[$20] - Submit and Register disable button color is not correct


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Open the URL

- Login as a valid user

- Go to /challenges/30058529

- Check the Submit and Register disable buttons

Actual Result(s)
-----------------------------------------------------------------
- Button color is BLUE

Expected Result(s)
-----------------------------------------------------------------
- Button color must Gray

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![3](https://user-images.githubusercontent.com/4592712/29668023-4f752d36-88fc-11e7-8513-d870aea38c66.jpg)

sdgun:
![3 0](https://user-images.githubusercontent.com/4592712/29668289-40857bfe-88fd-11e7-8c87-d7328cd03f7a.jpg)

birdofpreyru:
I accept it (1 point), as it is indeed different from design specs.
Though, @vic-topcoder wanna double check with you, should we make it gray or leave pale blue? Originally, it was done pale blue, because we took track-dependent color accents too seriously, so design / dev / datasci pages had these buttons blue / green / orange when active, and pale versions of these colors when disabled.

vic-topcoder:
Correct: From original design point the cases are:

- Not registered - all buttons but Register are disabled (gray style by default).
- Registered - can unregister (danger button), and primary Submit
- Submitted - can unregister (danger), Submit and Manage submissions (primary)

The latest styles in https://github.com/appirio-tech/tc-ui/blob/feature/connectv2/src/styles/_buttons.scss take care of all those states - reuse the styling please.

birdofpreyru:
### Expected Fix
Ok, should be fixed according to Vic's comment above. Though, rather than adding those styles he mentioned directly to button elements, be sure just to modify themes of standard buttons in `/src/shared/components/buttons` so that the result matches the specs

birdofpreyru:
@vic-topcoder Wanna double-check once more. In the challenge details style specs these two pages actually contradict each other:
[...]

PR #302, opened 2017-08-25
Issue #175

shubhendusaurabh:
fixed #175 styles for default button

### DIFF
--- a/src/shared/components/buttons/default.scss
+++ b/src/shared/components/buttons/default.scss
@@ -47,6 +47,8 @@
   color: $tc-gray-30;
   cursor: not-allowed;
   padding-top: 2px;
+  opacity: 0.3;
+  filter: grayscale(90%);
 }
 
 .link {


birdofpreyru:
@shubhendusaurabh Does not look right:
1. The resulting disabled buttons look a way too bright in the page, than in the specs, I think you put wrong colors there.
2. `src/shared/components/buttons/default.scss` is not a wrong style to edit. Actually, we have a demo / test page with available button styles: http://community-app.topcoder.com/examples/buttons/ changing `default.scss` you update mostly the styling of all default buttons in the app, which has no direct relation to these primary button of blue color.
3. Though, hold off, I have re-checked design specs, and actually, in UI styles page, the disabled state of primary (blue) button has pale-blue background. And gray buttons in the challenge details header actually have style of disabled secondary button. I want to double check this with Vic, and then update you on what should be done.

birdofpreyru:
@shubhendusaurabh Ok, there is the answer from Vic in the original issue ticket. I guess, you should just modify themes of standard buttons so that they have their disabled state defined by the code given by Vic.

shubhendusaurabh:
@birdofpreyru updated

Issue #294, opened 2017-08-25
[$20] - Submission count is not displayed in the tab


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges/30056022/?type=design

- Click on Submission tab and check the count in the tab header

Actual Result(s)
-----------------------------------------------------------------
- Submission count is not displayed in the tab

Expected Result(s)
-----------------------------------------------------------------
- Submission count must displayed in the tab

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![57 2](https://user-images.githubusercontent.com/4592712/29709946-085b9826-89ac-11e7-8653-b9a7ece3bda0.jpg)
![57 1](https://user-images.githubusercontent.com/4592712/29709947-089722ba-89ac-11e7-878f-f320331122ea.jpg)

birdofpreyru:
Accepted, 3 points.

PR #307, opened 2017-08-25
issue294 -iversonLv- Submission count is not displayed in the tab

iversonLv:
PR for issue #294

### DIFF
--- a/src/shared/components/challenge-detail/Header/TabSelector/index.jsx
+++ b/src/shared/components/challenge-detail/Header/TabSelector/index.jsx
@@ -21,6 +21,7 @@ export default function ChallengeViewSelector(props) {
     challenge,
     checkpointCount,
     numRegistrants,
+    numSubmissions,
     onSelectorClicked,
     selectedView,
     status,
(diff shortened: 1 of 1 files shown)


src/shared/components/challenge-detail/Header/TabSelector/index.jsx
```diff
@@ -61,8 +62,7 @@ export default function ChallengeViewSelector(props) {
           <a
             onClick={(e) => { e.preventDefault(); onSelectorClicked('SUBMISSIONS'); }}
             styleName={getSelectorStyle(selectedView, 'SUBMISSIONS', trackLower)}
-          >
-            SUBMISSIONS
+          >SUBMISSIONS {numSubmissions ? `(${numSubmissions})` : ''}
```
birdofpreyru: No need to make a check here, I believe: you have already specified in `ChallengeViewSelector.defaultProps` that default for `numSubmissions` is zero. So, even if no value is passed by parent component via props, it will be automatically set to 0 anyway, so you can just render `{numSubmissions}`. Please, change, and double-check that it works fine.

Issue #221, opened 2017-08-24
[$80] - 'TCO18' Tag in subheading of challenge details not consistent with design spec.


Kvm87:
Steps to Reproduce
1. Login to open http://community-app.topcoder.com/challenges/30058868/?type=design#winner
2. Observe 'TCO18' tag under heading title of the challenge
3. Also open http://community-app.topcoder.com/challenges/30058970/?type=develop
4. Observe 'TCO18' Tag at same location for this development challenge.
5. Also check 'Header' styleguide in design spec.

Screenshots : 
![image](https://user-images.githubusercontent.com/20595047/29673569-47145958-890d-11e7-8607-6245077f65a2.png)
![image](https://user-images.githubusercontent.com/20595047/29673580-4d45c596-890d-11e7-87e1-410ff4edf926.png)

Expected Result: As per design spec, the mentioned 'TCO' tag should be stylised in such a way that its background to be in light shade of color specific to the challenge type and letters to be written in true color.Hence, here 'TCO18' tag for design challenge should show light blue background with wording written in blue.

Current Result: In current system, for design challenges 'TCO18' tag is present in incorrect format.However, for development challenge it is coorect and consistent with design spec.

Browser : Chrome Version 59.0.3071.115 (Official Build) (64-bit)
OS : Windows 8.1

birdofpreyru:
Accepted, 1 point

### Expected Fix
I would like to make the fix in a fancy way: check the new implementation for buttons, themed using `react-css-themr` in `/src/shared/components/buttons`. These tags are, basically, button-like links with different styling, so we should just add additional themes to style that standard button component as tags, and ensure that proper theme is taken from `ThemeProvider`.

Feel free to ask for more details, if the brief note above does not look creal to you.

Colinh84:
@shubhendusaurabh Did you tag the wrong issue here, or did you really submit a fix for the issue I am assigned?

shubhendusaurabh:
@Colinh84 sorry my bad. I meant #211

PR #314, opened 2017-08-25
Issue 221

Colinh84:
https://github.com/topcoder-platform/community-app/issues/221


birdofpreyru:
@Colinh84 OK, it is a solid move in the right direction, but I want some changes:
1. Don't put tags staff into `/src/shared/components/buttons` make a separate `/src/shared/components/tags`. Please call default tag component just as `Tag` not `TagButton`; also don't call its stylesheet `tags.scss`, let it be just `default.scss`.
2. What you did technically looks right for the default tag (I am reviewing just looking into the code, have not launched it), but the default tag should be that gray one, as in the challenge listings.
3. For blue / green / orange tags you make separate stylesheets, similar to `primaryDesign.scss`, etc. stylesheets for buttons (pay attention, you don't have to copy / paste entire stylesheets, you can just import your tags `default.scss` into `primaryDesign.scss` and override only those few style rules that are different - this is completely similar to buttons.
4. To correctly style track dependent tags, you should just do `PrimaryTag` component, similar to `PrimaryButton`, and use context styling:
if you have declared
```
export const PrimaryTag =
  themr('PrimaryTag, primaryDesignTag)(GenericButton);
```
and you use it inside challenge details header as `<PrimaryTag ... />`
then to use proper style for that track you just go here: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/components/challenge-detail/themeFactory.js
and ensure that the proper theme is configured for each track for primary tags (again, it will be completely similar to the buttons related code, which is commented out there, because we finally did not need to make button colors track-dependent).
5. Finally, it will be great if you add to the repo a simple demo / test page for the tags you have implemented, similar to that one we have for buttons: http://community-app.topcoder.com/examples/buttons/

I'll bump the payment for this ticket by $50 for your further efforts. Though, these changes should not be that hard.

Colinh84:
@birdofpreyru I have implemented your requested changes, including an example page for the new Tag component and updated the jest unit tests. I rebased onto develop branch since there was a conflict with a change in the last 24 hours. With the rebase, I noticed that loading the main challenge listings page no longer works and there is an error `Uncaught (in promise) TypeError: Cannot read property 'filter' of null` in the console; I retested on the develop branch without my changes and got the same result, so it seems there is a bad commit on develop branch dealing with filters in the last day or so..

twicoder:
PR #421 fixed this


On 08/31/2017 23:01, Colin Hunt wrote:

@birdofpreyru I have implemented your requested changes, including an example page for the new Tag component and updated the jest unit tests. I rebased onto develop branch since there was a conflict with a change in the last 24 hours. With the rebase, I noticed that loading the main challenge listings page no longer works and there is an error Uncaught (in promise) TypeError: Cannot read property 'filter' of null in the console; I retested on the develop branch without my changes and got the same result, so it seems there is a bad commit on develop branch dealing with filters in the last day or so..

birdofpreyru:
@twicoder @Colinh84 Yeah, somehow I have not noticed this small bug during the merge.
Though, it is better to wrap all that code block into a single `if (query) { ... }`, so I'll close #421 and push that correction myself right now.

birdofpreyru:
@Colinh84 OK, accepted, overally good job. Though, quite many linter errors failing `npm test` (I've cleaned them up myself, while testing your PR). You should definitely setup ESLint and Stylelint plugins for your IDE.

Also right after merging I've noted some deviations between design specs and your styling, but I'll also fix myself, not a big deal.

Issue #303, opened 2017-08-25
[$30] - '0' is missing (or "min" should be erased) at remaining time


Vetteru:
## Steps to Reproduce
- access to http://community-app.topcoder.com/challenges (see challenges that end time is now)

## Actual Result(s)
- 5d min

## Expected Result(s)
- 5d 0min or 5d

## Environment
```
Device(s): Laptop
Operating System: MacOS 10.12.6
Browser(s): Google Chrome 60.0.3112.101 (Official Build) (64-bit)
```

## Image/Video/JS Log/Console Log (If not attached here, Please check the comments section)
![screen shot 2017-08-25 at 22 00 26](https://user-images.githubusercontent.com/6219446/29715299-b8a33b82-89e1-11e7-9025-90c96d93d420.png)

birdofpreyru:
Accepted, 3 points.
That text should be properly formatted. If it is more than one day, we should show Xd YYh (i.e. days and hours); once it goes under one day, we should show XXh YYmin (i.e. hours + minutes), and when it goes under 1 hour, then minutes and seconds.

twicoder:
PR #320

PR #320, opened 2017-08-25
Fix issue #303

twicoder:


birdofpreyru:
1. Have not actually tried to launch your fix, just looking into the code, I see, that it probably makes the trick right, but, please, re-write it to make necessary date/time formatting with `momentjs` (https://momentjs.com/docs/). I believe, it will be a way more compact, and a way less error-prone.

2. Also pay attention that these changes break some snapshot-based unit tests - should be fixed.

birdofpreyru:
@twicoder

twicoder:
@birdofpreyru  I have changed the code but can't find a proper method to format duration.

birdofpreyru:
@twicoder eh... https://momentjs.com/docs/#/displaying/format/ doesn't it cover our needs here?

twicoder:
@birdofpreyru The expected fix is show duration as “??d ??h” or “??h ??mim” or “??min ??s”, and I find the format function of moment can’t cover this well.

Issue #196, opened 2017-08-24
[$20] - [Bug Hunt - NEW BUG] Filter with duplicate name


twicoder:
1) Click save filter, but click it many times and fast, you will see duplicate items
<img width="643" alt="2017-08-24 10 09 27" src="https://user-images.githubusercontent.com/3046795/29670396-1a2d2e5e-8919-11e7-8bd6-841a4a72a996.png">

birdofpreyru:
Accepted, 3 points
Yeah, we should ensure such corner case is handled properly.

daft300punk:
Does saving filters in dev not work? I'm getting cors issue saying `Fetch API cannot load https://lc1-user-settings-service.herokuapp.com/saved-searches.`

As for the approach, I think we need to disable the save filter button unless success/failure response of filter getting saved comes back from the server. Is that fine?

birdofpreyru:
@daft300punk Yes, the api for saving filters exists only in prod. The approach to disable the button until the result, looks fine to me.

daft300punk:
I've a PR ready, but not tested on prod. Do we have a way to test things on prod? When I open challenge-details prod pages mentioned on localhost:3000 i.e. the page that lists various links, I get some strange promise rejection errors.

daft300punk:
Check out the PR - https://github.com/topcoder-platform/community-app/pull/322

birdofpreyru:
@daft300punk Yeah, to test locally against prod backend you run as `NODE_ENV=production PORT=80 npm run dev`, and browse the site as `local.topcoder.com`. Sure, you should first permit the app to listen on the port 80, there are some hints in README.md on how to configure it on Ubuntu.

daft300punk:
Updated PR - https://github.com/topcoder-platform/community-app/pull/322

PR #322, opened 2017-08-25
Filter with duplicate name #196

daft300punk:
This is pretty straight forward. Just a new redux state for storing whether or not a new filter is currently being saved. If it is being saved, we disable the save filter button.

### DIFF
--- a/src/shared/actions/challenge-listing/sidebar.js
+++ b/src/shared/actions/challenge-listing/sidebar.js
@@ -138,6 +138,8 @@ export default createActions({
 
       SAVE_FILTER: saveFilter,
 
+      SAVING_FILTER_BEGIN: _.noop,
+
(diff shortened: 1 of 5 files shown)


src/shared/actions/challenge-listing/sidebar.js
```diff
@@ -138,6 +138,8 @@ export default createActions({
 
       SAVE_FILTER: saveFilter,
 
+      SAVING_FILTER_BEGIN: _.noop,
```
birdofpreyru: @daft300punk Have not tried, though, looking through the code, it looks fine. But... please rename these actions into `SAVE_FILTER_INIT` and `SAVE_FILTER_DONE` to be consistent with naming of other action pairs for async actions.

Issue #176, opened 2017-08-24
[$30] - [Bug Hunt - NEW BUG]Web page flash when no items choosen


twicoder:
1) Unselect all types of contests (design, develepment, data science)
2) It's expected to see no item, but we see many empty items
3) And the web page is flash again and again after several seconds
<img width="1643" alt="2017-08-24 9 19 17" src="https://user-images.githubusercontent.com/3046795/29668372-85030ffc-8912-11e7-9541-38b29a05da66.png">

birdofpreyru:
Accepted, 3 points.
### Expected Fix
Indeed, at the moment the frontend code does not know that such (or, probably, some others as well) combination of filters cannot satisfy any challenge in principle. So, the frontend just keeps trying to find past challenges matching these fitlers. We need some elegant solution to avoid it.

ordinaryname:
Can you assign me to this issue please?

birdofpreyru:
@ordinaryname Added you to collaborators. Can you check, whether READ access is enough to self-assign?

ordinaryname:
I cannot self-assign

birdofpreyru:
@ordinaryname and now?

ordinaryname:
Fixed issue with Pull Request #323

birdofpreyru:
A simple and a bit hacky solution has been merged in by myself.
Though, with this solution our frontend still does one cycle of search of past challenges when all track filters are unset. We are still interested in a better fix (i.e. when we set a combination of filters which cannot match anything at all, we should not start the regular search procedure for past challenges). And it should not focus to match on the tracks switches, in a sense that it should be straightforward to add other checks of filters state, when this workaround should be used.

PR #323, opened 2017-08-25
Fixed issue #176

ordinaryname:
When all types of contests (design, develepment, data science) are unselected the page should load all contests, but instead it loads it tries to search for and load no contests. It tries again over and over for a few minutes and eventually stops trying.

I added an if statement in the 'removeTracks' function to delete the 'tracks' field similar to what was done in the 'addTracks' function when all tracks are selected. In the comments in the code it said that selecting all tracks is the same as having no tracks field. So selecting no tracks should also be the same.

When all tracks are removed the tracks field becomes an empty array which seems to cause the problem. Deleting the tracks array (field) makes the other functions behave as they should and return all challenges.

local.topcoder.com/challenges?

### DIFF
--- a/src/shared/utils/challenge-listing/filter.js
+++ b/src/shared/utils/challenge-listing/filter.js
@@ -326,6 +326,10 @@ export function removeTrack(state, track) {
     });
   }
   delete res.tracks[track];
+  /*As was stated in the 'addTrack' function, selecting all tracks is the same as having
+   no tracks field. So selecting no tracks should also be the same. Removing this line
+   causes no challenges to be returned and flickering of the empty challenge listings.*/ 
+  if (_.isEmpty(res.tracks)) delete res.tracks;
   return res;
 }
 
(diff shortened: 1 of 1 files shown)


birdofpreyru:
@ordinaryname No, it leads to incorrect logic: we disable all tracks, but we see all challenges. I would expect, that disabling all tracks shows no challenges. An easy way to achieve that is to modify https://github.com/topcoder-platform/community-app/blob/develop/src/shared/utils/challenge-listing/filter.js#L302
add there a check if `tracks` filter is present, but empty return as the result the object
```
{
  groupIds: '0',
}
```
With this, when all tracks are disabled, the backend will be called with backend filter for challenges from group `0`, but there is no such group, so we'll get 0 challenges in the response immediately > no flickering. Just double-check, that this actually works :)

ordinaryname:
It works. I submitted a new pull request  ( #415 ) starting fresh with a new clone of the develop branch and adding the check of the tracks filter.

Issue #187, opened 2017-08-24
[$40] - [Bug Hunt - NEW BUG]n.subtracks.join is not a function Challenge-list.js:11


twicoder:
1) Try to open page: 
http://community-app.topcoder.com/challenges?filter[tags][0]=Java&filter[subtracks][0]=Architecture&filter[subtracks][1]=Banners/Icons&filter[subtracks][2]=Conceptualization&filter[subtracks][3]=Design&filter[subtracks][4]=First2Finish&filter[subtracks][5]=Logo%20Design&filter[subtracks][6]=RIA%20Build%20Competition&filter[subtracks][7]=Test%20Suites&filter[subtracks][8]=Wireframes&filter[subtracks][9]=Widget%20or%20Mobile%20Screen%20Design&filter[subtracks][10]=Test%20Scenarios&filter[subtracks][11]=Studio%20Other&filter[subtracks][12]=Print/Presentation&filter[subtracks][13]=Marathon%20Match&filter[subtracks][14]=Front-End%20Flash&filter[subtracks][15]=Development&filter[subtracks][16]=Design%20First2Finish&filter[subtracks][17]=Copilot%20Posting&filter[subtracks][18]=Bug%20Hunt&filter[subtracks][19]=Code&filter[subtracks][20]=Application%20Front-End%20Design&filter[subtracks][21]=Assembly%20Competition&filter[subtracks][22]=Content%20Creation&filter[subtracks][23]=Idea%20Generation&filter[subtracks][24]=Web%20Design&filter[subtracks][25]=Specification&filter[subtracks][26]=UI%20Prototype%20Competition&filter[tracks][design]=true&filter[tracks][develop]=true&filter[startDate]=2017-08-01T04:00:00.000Z&filter[endDate]=2017-08-05T04:00:00.000Z

2) Page can't show items and there is console error:
<img width="1657" alt="2017-08-24 9 49 43" src="https://user-images.githubusercontent.com/3046795/29669618-a493187c-8916-11e7-96a4-d4b901d09a3f.png">

birdofpreyru:
Accepted, 5 points.
I am not sure, whether this is a valid URL or not, from point of view of proper filters logic, but we surely should not allow a user to break the app by supplying incorrect query params in URL. So, it is necessary to check, what is the issue, and fix it properly.

birdofpreyru:
Pay attention to #195 - should be also fixed as a part of this

elfman:
assign to myself

PR #327, opened 2017-08-26
Fix issue #187

elfman:


src/shared/components/challenge-listing/Filters/FiltersPanel/index.jsx
```diff
@@ -53,6 +53,17 @@ export default function FiltersPanel({
 
   const mapOps = item => ({ label: item, value: item });
 
+  let filterStateTags = null;
+  if (filterState.tags) {
+    filterStateTags = Array.isArray(filterState.tags) ? filterState.tags.join(',') :
+      Object.keys(filterState.tags).map(key => filterState.tags[key]).join(',');
+  }
```
birdofpreyru: 1. For such checks we prefer to use lodash functions.
2. Is really possible that we can get either array either object here? It looks strange.

src/shared/utils/challenge-listing/filter.js
```diff
@@ -113,7 +113,9 @@ function filterBySubtracks(challenge, state) {
    * it probably does not work in all cases. It should be double-checked,
    * why challenge subtracks in challenge objects are different from those
    * return from the API as the list of possible subtracks. */
-  const filterSubtracks = state.subtracks.map(item =>
+  const subtracks = Array.isArray(state.subtracks) ? state.subtracks :
+    Object.keys(state.subtracks).map(key => state.subtracks[key]);
+  const filterSubtracks = subtracks.map(item =>
```
birdofpreyru: Especially here. It is just documented in the module's head comment that `tags` and `subtracks` must be arrays. So if it gets called with object, it is an error to be fixed somewhere else.

elfman:
@birdofpreyru I found another way to solve this bug, please check again

src/server/server.js
```diff
@@ -34,6 +35,8 @@ const app = express();
  * fix. */
 global.atob = atob;
 
+app.set('query parser', str => qs.parse(str, { arrayLimit: 1000 }));
```
birdofpreyru: @elfman It looks like you found the correct reason behind the issue, but I don't like this way solving it. The reason `qs` limits possible array lengths and turns array elements with indices larger than 20 into objects is the following: say you set limit 1000, then it is legit to send query param x[999]=... and the backend will have to create an array with 1000 elements for this param. I guess, it opens a very easy way to spam the server, if somebody wants to: you just send in a query like x[999]=a&y[999]=b&... and server will have lots of fun creating a big array for each of params.
So, I guess, we better just account in the reducer that long arrays are actually parsed as objects.

elfman:
Updated

Issue #284, opened 2017-08-25
[$20] - Stalled challenges displayed in 'Open for registration' section


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges

- In the 'Open for registration' section > Sort by: Time to Register

- Check the Stalled challenges

Actual Result(s)
-----------------------------------------------------------------
- Stalled challenges displayed in 'Open for registration' section

Expected Result(s)
-----------------------------------------------------------------
- Must omit the Stalled challenges from the Open for registration

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![53](https://user-images.githubusercontent.com/4592712/29704270-b713ef28-8996-11e7-91bb-a3c4e0b97095.jpg)

birdofpreyru:
Accepted, 3 points.

elfman:
The bug is caused by the data from backend, it contains some mistake, the registration time has passed, but the status of registration phase is still open.
The filter used phase state to judge if the registration is open, now I change it to use registration time. But I still think the backend should fix this.

birdofpreyru:
@elfman This fixes the issue for MMs, but not for F2F. I have feeling that I've already saw a fix for F2F case somewhere during this Bug Bash, but I am really a bit lost now, so I don't remember, was it already merged-in and broken again, or is it still pending the merge. Anyway, I don't accept this at least until I find the fix for F2F.

PR #331, opened 2017-08-26
Fix issue #284

elfman:


birdofpreyru:
@elfman Does not look working with MM for me. Probably, because MM challenges don't have registration dates, as the regular challenges do, so your updated code is just skipped for them, and they are still shown as stalled.

elfman:
I am really confused with the challenge data, why challenge's `status` has been set to `PAST`, but the field `registrationOpen` is still true, what is true state of this challenge?
145aec03-c19b-4e3d-8342-d783d3af471a

elfman:
In some regular challenges, the registration date has past, why the value of `registrationOpen` is still `Yes`. Which filed should I use?
691b8609-9b37-4615-ba2e-e6ce69206496

birdofpreyru:
@elfman Please double-check the code related to fetching challenges. I believe, what happens is that `registrationOpen` field is not actually fetched from the backend: regular challenges and MMs are fetched from different api endpoints by challenge service, and then they are passed through a normalization function, which ensures that these objects have the same format (by adding missing fields, and doing any other modifications, necessary for the rest of the code to handle these objects in the same way). I guess, `registrationOpen` field is added there, and there might be an error with setting it for some challenges (probably, it is actually not used anymore around the code, then should be removed).

elfman:
@birdofpreyru I checked the origin data before `normalizeChallenge`, and I don't understand why the `scheduledEndTime` of `Registration` is different from `registrationEndDate`
67b4518f-8d9a-4bb4-af62-566732427047

birdofpreyru:
@elfman Please re-check with production backend. Dev backend has problems with sync between v2 and v3 APIs, probably this is the reason.

elfman:
@birdofpreyru Updated, please check again.

src/shared/services/challenges.js
```diff
@@ -79,7 +79,8 @@ export function normalizeMarathonMatch(challenge, username) {
     numSubmissions: challenge.userIds ? challenge.userIds.length : 0,
     platforms: '',
     prizes: [0],
-    registrationOpen: endTimestamp > Date.now() ? 'Yes' : 'No',
+    registrationOpen: endTimestamp > Date.now() &&
+      !(challenge.status && challenge.status === 'PAST') ? 'Yes' : 'No',
```
birdofpreyru: @elfman `!(challenge.status && challenge.status === 'PAST')` is equivalent to just `challenges.status !== 'PAST'`, which is surely more clean, and thus preffered code. Will update myself, though.

Issue #337, opened 2017-08-26
[$20] - Placeholder text of SearchBox is not in the middle


elfman:
## Steps to Reproduce
Goto http://community-app.topcoder.com/challenges/

## Actual Result(s)
placeholder text of SearchBox is not in the middle

## Expected Result(s)
placeholder text aligned in the middle

## Environment
```
Device(s): Mac mini & iPad Air2
Operating System: MacOS 10.11.6 & iOS 9.3.2
Browser(s): Safari 10.1.2
```

#### Mac
![Mac Safari](https://user-images.githubusercontent.com/948001/29740061-763a6992-8a80-11e7-99c8-c89cecdd2ff7.png)

#### iPad
![iPad Safari](https://user-images.githubusercontent.com/948001/29740051-43c3c0d0-8a80-11e7-8331-ad84fe40ece0.png)

birdofpreyru:
Accepted, 3 points.
I believe, this happens only on some browsers (e.g. works fine in Chrome on Ubuntu). We should check related style to ensure it is not broken on main browsers (note that we have autoprefixer setup in the app, so probably it is worth to check here, why it did not help in this case).

elfman:
@birdofpreyru I just notice that the Topcoder Header is missing in tablet (see the second picture in the issue), is it a bug or a feature?

elfman:
@birdofpreyru what's wrong with backend now? I can't open challenge detail page now, it works several hours ago. I see 504 error and `No 'Access-Control-Allow-Origin' header'` in console

elfman:
@birdofpreyru I made Topcoder Header displayed on iPad, and I found some bug on it.
1. when iPad is landscape, the page use the design of PC, when click on the menu item, subMenu opened, but I can't find a way to close the submenu. (This is not the styling problem, I think I should open another issue for this bug.)
2. the search icon at top-right is not at the right position.
3. the page use the design for PC, but the Topcoder Footer use the design for mobile.

#### Header on iPad
![topcoder-header](https://user-images.githubusercontent.com/948001/29742519-47d8e82a-8ab3-11e7-90ed-b6c35c1476b3.png)

#### Footer on iPad
![topcoder-footer](https://user-images.githubusercontent.com/948001/29742521-4fc1e5fa-8ab3-11e7-802e-b5fa104518de.png)

birdofpreyru:
@elfman Yeah, please log it as a sepate issue. It is a good catch about desktop header in landscape view on tablet.

[...]

Issue #339, opened 2017-08-26
[$60] - Can not close SubMenu of TopCoder Head[er on tablet


elfman:
Steps to Reproduce
-----------------------------------------------------------------

- iPad at landscape, Go to 'local.topcoder-dev.com:3000/challenges' (currently the Topcoder Header is not shown on public server, so I use local dev, it will be fixed in #337)

- Tap on Menu item, SubMenu appears

- Tap on clean area of the page, the subMenu doesn't close

Actual Result(s)
-----------------------------------------------------------------
- Can not close SubMenu in landscape view on iPad

Expected Result(s)
-----------------------------------------------------------------
- When tap on empty area of page, close the submenu.

Environment
-----------------------------------------------------------------
- **Device(s):** iPad Air2
- **Operating System:** iOS 9.3.2
- **Browser(s):** Safari

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section)

birdofpreyru:
Yeah, we should update header's code, to ensure that it works with clicks in addition to hovering. I.e. clicking on the main menu item's opens submenu, clicking outside of the submenu items, when the submenu is opened, closes it.

Issue #223, opened 2017-08-24
[$60] - Menu stays forever when user move the mouse down over COMPETE/LEARN/COMMUNITY


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges/30058991/?type=develop

- Click Filter and apply any filter

- While filtering move the mouse down over COMPETE/LEARN/COMMUNITY and quickly scroll the page down

- Now scroll up you can see menu is activated and click on any empty location

Actual Result(s)
-----------------------------------------------------------------
- Menu stays forever when user move the mouse down over COMPETE/LEARN/COMMUNITY

Expected Result(s)
-----------------------------------------------------------------
- Must close the menu when user tap on a empty space

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

sdgun:
![2017-08-25_5-05-05](https://user-images.githubusercontent.com/4592712/29693412-0d619dba-8953-11e7-97f3-194d6fbab677.jpg)

birdofpreyru:
@sdgun Not sure about this one. When I click on empty space, then move mouse out of the menu, it still collapses. Do you mean, that we should close the menu immediately if we click on its empty part (dark gray zone)? It does not look that obvious to me. @vic-topcoder ?

vic-topcoder:
That's a mobile interaction you're describing: on mobile to invoke the menu one has to tap to simulate hover state. On desktop it should not happen. I cannot reproduce this on my side. If your mouse is outside the menu, it should collapse. Always-persistent menu is a bug

sdgun:
@vic-topcoder @birdofpreyru Please give me sometime I will comment the exact steps. I can reproduce this 8/10 times

sdgun:
@birdofpreyru Please check the attached video;

https://www.screencast.com/t/lzsEbOULW

birdofpreyru:
@sdgun Ahhh... I see. It looks the actual reason is that you scroll at the same time when the menu is being opened, so this interfere with JS code taking care about the menu. Accepted, 5 points.

### Expected Fix
The logic behind sub-menu expansion / collapsing should be revised to ensure that scrolling does not break it.

PR #345, opened 2017-08-27
Fix issue #337, #339, #223

elfman:
#339 rely on #337 be fixed, so I rise PR for them together

### DIFF
--- a/src/shared/actions/topcoder_header.js
+++ b/src/shared/actions/topcoder_header.js
@@ -9,24 +9,28 @@ import { createActions } from 'redux-actions';
  * Payload creator for OPEN_MENU action.
  * @param {Object} menu Menu to open.
  * @param {Object} triggerNode HTML node which has triggered the action.
+ * @param {Boolean} triggeredByTouch whether this action is triggered by touch event.
  *  We gonna store to the state its on-screen layout position, so that
  *  when mouse cursor moves back and forth between that node and the
  *  opened menu, we can check it and keep the menu open.
  */
-function openMenu(menu, triggerNode) {
+function openMenu(menu, triggerNode, triggeredByTouch) {
   return ({
     menu,
+    node: triggerNode.dataset.menu,
     trigger: _.toPlainObject(triggerNode.getBoundingClientRect()),
+    isMobile: triggeredByTouch,
   });
 }
 
 /**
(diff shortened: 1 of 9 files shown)


birdofpreyru:
@elfman Had no chance to test it carefully, though, during a brief try I've noted the following problem:
- If I open desktop kind of menu on tablet (by tapping on, say, `Learn`) and then tap on the white part of the header (i.e. the same element where `Compete`, `Learn` and `Community` buttons are, just outside of them, and other elements, menu is not closed and there are some error in console.

elfman:
Fixed, please test again.

elfman:
issue #223 and #339 are both related with TopcoderHeader, so I put it in this PR too.
Please take it account when testing.

birdofpreyru:
@elfman Wow :) This are big changes, and they seem to be open.

There is just one more enhancement that would be handy: now if I use switch browser into touch device emulation mode, use touches to play with the menu, then switch to the normal mode and try to interact with the menu neither hovering nor clicks work anymore until I reload the app. I understand, that it is not an issue for production use, as user, in the same session, will not switch like this. Though, for dev/testing routine, it would be super-cool to fix it, so that such switches do not break anything. Also, I guess, it should be just a small modification somewhere, as there is no real need to disable all hover-related logic once you have opened a menu with touch, right?

elfman:
MouseEnter & MouseLeave Event can also be triggered in mobile device. But their event logic is not fit to mobile device and can't work with touch event. For example, I tap on a menu button when menu opened, the menu should be closed as touch event do so, but the MouseEnter will also be trigger, and it will open the menu again, then the menu looks like nothing happened.
so I disable mouseenter and mouseleave when touch event triggered.
If you want to make it work when switching emulation mode, how about reset the `isMobile` state in `onLayout` when `width` changed?

birdofpreyru:
@elfman No, I don't like listening the width change, as it probably will slow down window resizing (which is already not that fast). Can we just re-enable `mouseenter` and `mouseleave` once the menu is closed? I believe, it is already closed automatically if I switch between touch device and desktop in the way described above. And having them enabled when the menu is closed is not a problem for us?

elfman:
That's a good idea, I'll do it

elfman:
@birdofpreyru Updated.

Issue #282, opened 2017-08-25
[$50] - Sort by: Time to submit will sort the list by Time to register


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Login as a valid user

- Go to http://community-app.topcoder.com/challenges

- Sort by: Time to submit in 'My Challenges'

- Check the sorted list

Actual Result(s)
-----------------------------------------------------------------
- Sort by: Time to submit will sort the list by Time to register

Expected Result(s)
-----------------------------------------------------------------
- Sort by: Time to submit must sort the list by Time to Submit

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 64bit
- **Browser(s):** Chrome 60.0.3112.101 (Official Build) (64-bit)

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------

birdofpreyru:
Accepted, 3 points. I am not actually sure, in what order does it sort challenges now, but it looks working definitely wrong. Should be fixed.

birdofpreyru:
Also pay attention to #286. Seems to be the same problem, so both should be fixed here.

Colinh84:
@birdofpreyru Need some clarification on this. It appears the results ARE actually being sorted by time to submit correctly, though for "My Challenges" the results are listed in ASC order from least -> most time remaining and "Open for Registration" listings are sorted in DESC order from most -> least time remaining. See screenshot of where time to submit is listed:
![sort-order](https://user-images.githubusercontent.com/2287621/29736748-2041d2ea-89b9-11e7-9363-1a7279efea1b.png)

What is the correct order to sort these categories by (ASC or DESC time remaining)? Other than that order difference, the challenges are sorting correctly except for one stalled challenge shown at the top of the "open for registration" category.

birdofpreyru:
@Colinh84 I better double-check tomorrow. But, in principle, I would expect that sorting by sumbmission time, would sort challenges by time left till the next submission deadline, from the first one to the last. In case of dev. challenges we have just one submission deadline, so it is clear; in case of design ones, we take as submission deadline the least of the times till checkpoint submission, and till the final submissition, and sort these from soonest, to the farthest.

Colinh84:
[...]

PR #357, opened 2017-08-27
Issue 282: Fixed challenge listing sorting by next submission or checkpoint date

Colinh84:
https://github.com/topcoder-platform/community-app/issues/282

### DIFF
--- a/src/shared/utils/challenge-listing/buckets.js
+++ b/src/shared/utils/challenge-listing/buckets.js
@@ -40,7 +40,7 @@ export function getBuckets(userHandle) {
       name: 'My Challenges',
       sorts: [
         SORTS.MOST_RECENT,
-        SORTS.PHASE_END_TIME,
+        SORTS.TIME_TO_SUBMIT,
         SORTS.NUM_REGISTRANTS,
         SORTS.NUM_SUBMISSIONS,
         SORTS.PRIZE_HIGH_TO_LOW,
@@ -58,7 +58,7 @@ export function getBuckets(userHandle) {
       sorts: [
         SORTS.MOST_RECENT,
         SORTS.TIME_TO_REGISTER,
-        SORTS.PHASE_END_TIME,
+        SORTS.TIME_TO_SUBMIT,
         SORTS.NUM_REGISTRANTS,
         SORTS.NUM_SUBMISSIONS,
         SORTS.PRIZE_HIGH_TO_LOW,
--- a/src/shared/utils/challenge-listing/buckets.js
+++ b/src/shared/utils/challenge-listing/buckets.js
@@ -40,7 +40,7 @@ export function getBuckets(userHandle) {
       name: 'My Challenges',
       sorts: [
         SORTS.MOST_RECENT,
-        SORTS.TIME_TO_SUBMIT,
+        SORTS.PHASE_END_TIME,
         SORTS.NUM_REGISTRANTS,
         SORTS.NUM_SUBMISSIONS,
         SORTS.PRIZE_HIGH_TO_LOW,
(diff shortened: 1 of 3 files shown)


src/shared/utils/challenge-listing/buckets.js
```diff
@@ -40,7 +40,7 @@ export function getBuckets(userHandle) {
       name: 'My Challenges',
       sorts: [
         SORTS.MOST_RECENT,
-        SORTS.TIME_TO_SUBMIT,
```
birdofpreyru: Not sure, why do we need this change (see also further comments).

src/shared/utils/challenge-listing/filter.js
```diff
@@ -73,6 +73,9 @@ function filterByGroupIds(challenge, state) {
 function filterByRegistrationOpen(challenge, state) {
   if (_.isUndefined(state.registrationOpen)) return true;
   const isRegOpen = () => {
+    if ('registrationOpen' in challenge) {
```
birdofpreyru: No, please do such checks just as `if (challenge.registrationOpen) {`.
I am also not sure, why do we need to touch this filter at all?

src/shared/utils/challenge-listing/sort.js
```diff
@@ -34,7 +34,18 @@ export default {
     name: '# of submissions',
   },
   [SORTS.PHASE_END_TIME]: {
-    func: (a, b) => a.currentPhaseRemainingTime - b.currentPhaseRemainingTime,
+    func: (a, b) => {
```
birdofpreyru: I don't competely follow, why do you modify this sort function? I believe, the original code was correct, wasn't it? It trully compared the time till the phase end (which could be any phase, not only checkpoint and submission). The only thing that wasn't good was the measleading name of this sort.

On the other hand, the code itself seems fine logically, just had to be put into [SORTS.TIME_TO_SUBMIT]

src/shared/utils/challenge-listing/sort.js
```diff
@@ -34,7 +34,18 @@ export default {
...
+      function nextSubEndDate(o) {
+        if (("checkpointSubmissionEndDate" in o) && moment(o.checkpointSubmissionEndDate).isAfter()) {
```
birdofpreyru: The first part of condition should be just `o.checkpointSubmissionEndDate && ...`

src/shared/utils/challenge-listing/filter.js
Colinh84: @birdofpreyru This filter needed to be changed because the "open registration" challenge listing included results for stalled challenges which were closed a long time ago, but still appeared because they had a "status' key set to ACTIVE. There is another key in all challenge results that was not being checked in the filter called "registrationOpen", and that key was properly set to "No" on these stalled challenges, so doing this extra check here removes these old stalled, closed challenges from the open registration listings.

src/shared/utils/challenge-listing/buckets.js
Colinh84: @birdofpreyru The "open for registration" listing was already using SORTS.PHASE_END_TIME, which is the sort that I modified to fix the sort bug; since "my challenges" should be sorted the same way, I just updated it to use the same sort. The SORTS.TIME_TO_SUBMIT sort only sorts by final submission end time, while SORTS.PHASE_END_TIME sorts by end of next submission phase, including next checkpoints.

src/shared/utils/challenge-listing/sort.js
Colinh84: @birdofpreyru This was the main source of the bug. The original sort here was actually not sorting anything and returned the results unchanged from the server, because the "currentPhaseRemainingTime" is a key from the V2 API which is no longer in use for grabbing challenge listings and is not present in the V3 API. None of the challenge results contain this key anymore. This SORT.PHASE_END_TIME was the sort that was being used for "time to submit" in open for registration listings, was labeled as such and was used nowhere else, which is why I modified this one.

Colinh84:
@birdofpreyru I updated the branch with requested changes. Also see my previous replies to code comments.

Issue #346, opened 2017-08-27
[$20] - TopcoderHeader mobile menu triggered when click on JOIN/LOG IN Button


elfman:
## Steps to Reproduce
- access to http://community-app.topcoder.com/challenges/30058637 in portrait view on iPad
- Tap the JOIN or LOG IN Button

## Actual Result(s)
- mobile menu is opened before redirect to other page

## Expected Result(s)
- redirect to other page without other response

## Environment
```
Device(s): iPad Air2
Operating System: iOS 9.3.2
Browser(s): Safari
```

birdofpreyru:
Accepted, 3 points

PR #369, opened 2017-08-28
Fix issue #346

elfman:


src/shared/components/TopcoderHeader/mobile/Header/index.jsx
```diff
@@ -19,7 +19,11 @@ export default function Header({
   return (
     <div styleName="header-wrapper">
       <div
-        onClick={open}
+        onClick={({ target }) => {
+          if (target.nodeName !== 'A' || !target.getAttribute('href')) {
```
birdofpreyru: @elfman This looks goofy to me. I guess, it is better to add some code to the Log In / Out buttons in question, to ensure that when they are clicked, the events are not propagated to the menu header under them

elfman:
Updated, please checkout again.

PR #401, opened 2017-08-29
Fix Issue 376

twicoder:
1) Implemented as suggested

2) Because expected UI is changed, I have to update the test case file __tests__/shared/components/challenge-listing/Filters/__snapshots__/ChallengeFilters.jsx.snap to pass the 'npm test'


birdofpreyru:
@twicoder One small problem: the count is visible when we got 0 active filters - should not be displayed (the same as it is now in desktop view).

twicoder:
@birdofpreyru  Changed as suggestion, thanks!

PR #407, opened 2017-08-29
Issue 375

twicoder:
Change the numberOfMonths of daterangepicker from 2 to 1.


birdofpreyru:
@twicoder Your change affects both mobile and desktop date pickers. While for mobile this fixes the problem, please update to keep 2 months in the desktop case.

Also `npm test` is broken. Please fix.

twicoder:
@birdofpreyru  changed the logic to show datepicker with 2 month when desktop view and 1 month when mobile view.

Issue #287, opened 2017-08-25
[$20] - for few second text shown insted of profile image 


karthikbecse:
Steps to Reproduce

   
  - Goto http://community-app.topcoder.com/challenges/
  - Scroll down to past challenges 
  - move cursor on any winners(without profile image )
  - observe  

Actual Result(s)
- for few second text shown insted of profile image 

Expected Result(s)
- Should show the image properly 
 
  
Environment

    Device(s): Laptop
    Operating System: Windows 8.1  64bit
    Browser(s): FF 54.0.1

Image/Video/JS Log/Console Log 
https://www.screencast.com/t/TSD7YvAUEly

birdofpreyru:
Accepted, 3 points.

daft300punk:
Is this a browser issue? I don't see it happening in chrome.

birdofpreyru:
@daft300punk It happens only for winners without profile picture, when the default "anonymous man" image is used as the avatar. Can you see it?

daft300punk:
No, I can only see it mozilla, not in chrome. I'm on windows.

birdofpreyru:
I see in Chrome / Ubuntu.

elfman:
@birdofpreyru I'm on Mac, this bug happen in Chrome, not in FF
It seems to be caused by lazy loading image in browser, some browser download the image only when it should be shown on the screen, in that duration there will be only text in the tooltip and the text will be shown on the place where avatar should be.
To fix this, I want to keep the space of avatar before the avatar image loaded. There will still be only text  in tooltip during loading , but the text won't jump when loading finished.
is that OK?

PR #409, opened 2017-08-29
Fix issue #287

elfman:
Fix issue #287

### DIFF
--- a/src/shared/components/Avatar/index.jsx
+++ b/src/shared/components/Avatar/index.jsx
@@ -20,13 +20,14 @@
 
 import PT from 'prop-types';
 import React from 'react';
+import _ from 'lodash';
 import { themr } from 'react-css-themr';
 import DefaultAvatar from '../../../assets/images/ico-user-default.svg';
 import defaultStyle from './style.scss';
 
-function Avatar({ theme, url }) {
+function Avatar({ theme, url, handleError }) {
   return url
-    ? <img alt="Avatar" src={url} className={theme.avatar} />
+    ? <img alt="Avatar" src={url} className={theme.avatar} onError={handleError} />
     : <DefaultAvatar className={theme.avatar} />;
 }
 
@@ -36,6 +37,7 @@ Avatar.defaultProps = {
     avatar: '',
   },
   url: null,
+  handleError: _.noop,
 };
 
 Avatar.propTypes = {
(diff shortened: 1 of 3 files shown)


birdofpreyru:
@elfman Nope, we can do better. The problem with this default avatar in the tooltip is that the logic to show it is really goofy: it always tries to load the photo, and if it fails due to 404 error, then it updates the URL to the default avatar in the error handler.
1. There should be no errors: if winner object has no photo URL - we just use the default avatar immediately; if he has absolute photo URL - we use it; if the URL is relative (like `/i/m/some-avatar.jpg`), it should be prepended with the correct domain (now, when I am running locally, it goes to `local.topcoder.com/i/m/some-avatar.jpg`, which is wrong, thus fails).
2. We have a standard component for rendering avatars: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/components/Avatar/index.jsx which basically takes care  about styling. For the usecase in the listing, I guess, just the default theme of that component is the same, so we should reuse that.

elfman:
@birdofpreyru Updated as your suggestion, except the theme of avatar. The default avatar theme is so small, I change it bigger

birdofpreyru:
@elfman Well, but do we really need the `onError` handler? I would expect we don't, cause if we compose avatar URLs in the correct way (adding the right base domain from configuration) then all URLs gonna be just correct.

birdofpreyru:
So, no need to modify avatar component then

elfman:
@birdofpreyru I have composed the URLs in line 36 of `UserAvatarTooltip/index.js`. Adding `handleError` is just wanna to show a default avatar when some network error happen, in case there would be empty area.

birdofpreyru:
@elfman Hmm... I prefer to remove error handlers. If the page fails due to network error, broken avatar images is the least of the problems, as most probably it will also break everything else. So, no need for such complication of the code.

elfman:
@birdofpreyru Updated, please check.

elfman:
@birdofpreyru Currently on `origin/develop`, `ChallengeList` does not work when there is no query string, this is not introduced by me

birdofpreyru:
@elfman Btw, have you seen my last post in the challenge forum? I don't know what is your TC nickname to be able to pay for the Bug Hunt part of the challenge.

elfman:
@birdofpreyru sorry, my TC nickname is `windyluo`

PR #415, opened 2017-08-30
Fixed issue #176

ordinaryname:
I removed the code from the other pull request and then added and tested the recommended check to see if the tracks field is present and empty. When there are no entries in the tracks field the 'groupIds' is set to '0'. There is no group 0 so 0 challenges are returned immediately, eliminating the flickering.


birdofpreyru:
@ordinaryname No, this is not done correctly, and does not work. As the correct fix is really easy, I have to reject your efforts, and just do it myself. Sorry.

PR #472, opened 2017-09-18
Change margin/padding in banner.scss. These classes are used only at …

sushilshinde:
Fix for #471 

Change margin/padding in banner.scss. These classes are used only at one place in the whole project, so shouldn't affect other UI's.


sushilshinde:
Merge done!

PR #534, opened 2017-10-20
Update index.jsx

alamora:
Includes changes to copy: 
- add OperationCode
- Drop the VA
- change wording about college degrees


alamora:
Addresses ticket 535

Issue #587, opened 2017-11-14
[$30] - Update communities dropdown in header to be dynamic


ajefts:
Use the same logic we have in the challenge filter.

birdofpreyru:
### Expected Fix:
Just re-use the code and functionality being used by the challenge listing to load the listing of communities visible to the visitor/user, and use it to populate the selector in sub-community header.

zjuasmn:
PR: https://github.com/topcoder-platform/community-app/pull/615
hash:  6a82124

birdofpreyru:
@zjuasmn Great, accepted. I'll keep the PR open for some more time, as I want to double-check some things with TC team before I can merge it; but nothing to worry about on your side.

PR #615, opened 2017-11-28
#587 Update communities dropdown in header to be dynamic

zjuasmn:
There is no url in returned community list, so I assume the url for each community is`/community/${communityId}`.

### DIFF
--- a/src/shared/containers/tc-communities/CommunityStats.jsx
+++ b/src/shared/containers/tc-communities/CommunityStats.jsx
@@ -13,11 +13,13 @@ import shortid from 'shortid';
 
 import actions from 'actions/stats';
 import cActions from 'actions/challenge-listing';
+import communityActions from 'actions/tc-communities';
 import CommunityStats from 'components/tc-communities/CommunityStats';
 
 class CommunityStatsContainer extends React.Component {
   /* When container mounts we get / update related stats. */
   componentDidMount() {
+    this.props.getCommunityList(this.props.auth);
     this.props.getCommunityStats(this.props.community, this.props.challenges, this.props.token);
     this.props.getAllActiveChallenges(this.props.token);
   }
@@ -50,6 +52,7 @@ class CommunityStatsContainer extends React.Component {
 }
 
 CommunityStatsContainer.defaultProps = {
+  auth: {},
   community: {},
   stats: {},
   token: '',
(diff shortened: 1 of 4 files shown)


src/shared/containers/tc-communities/CommunityStats.jsx
```diff
@@ -13,11 +13,13 @@ import shortid from 'shortid';
```
birdofpreyru: The stuff you did in this file, looks correct, but does not work in practices, because not every community, and not every page inside communities, renders the CommunityStats container; thus in most of cases the list of communities visible to the user will never be loaded. You should move this logic into src/shared/containers/tc-communities/Header.jsx (yeah, there you'll have to add the class component wrapper to have `componentDidMount()` hook from where you can fire the get list action, when necessary).

src/shared/containers/tc-communities/Header.jsx
```diff
@@ -15,13 +15,23 @@ function mapStateToProps(state, ownProps) {
...
+  const selectorLabels = meta.communitySelector.map(({ label }) => label);
+  const communitySelector = meta.communitySelector.concat(
+    state.tcCommunities.list.filter(({ communityId, communityName }) =>
+      communityId !== meta.communityId && selectorLabels.indexOf(communityName) < 0)
+      .map(({ communityId, communityName }, i) => ({
+        value: (i + (meta.communitySelector || []).length).toString(),
+        label: communityName,
+        redirect: `/community/${communityId}`,
+      })),
+  );
```
birdofpreyru: This block of code looks correct, most probably, but it is placed in the wrong place: never ever create new objects inside `mapStateToProps()` to further pass them inside the component!!! If you do this, each time React re-evaluates the state, it will call this `mapStateToProps()`; it will create a new object for `communitySelector` (even when no state changes happened at all) and pass it inside React container; react will see that the object has been changed (it does shallow comparison) and will unnecessarily re-render entire sub-tree of components.

The correct way: pass entire `state.tcCommunities.list` inside the component, and then do any further filtering / mapping inside the render method, directly responsible for the dropdown. This way we avoid unnecessary re-renderings, and ensure good performance of the app.

zjuasmn:
@birdofpreyru please check my updated PR.

Issue #598, opened 2017-11-19
[$200] - JoinCommunity button should be updated to use the standard theming approach!


birdofpreyru:
### Expected Fix:
This component: https://github.com/topcoder-platform/community-app/tree/develop/src/shared/components/tc-communities/JoinCommunity
and its container: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/containers/tc-communities/JoinCommunity.jsx
should be updated to use the standard themable buttons (https://github.com/topcoder-platform/community-app/tree/develop/src/shared/components/buttons) inside them.

Then all communities where JoinButton is used should be checked and updated properly:
- All buttons inside the modals opened by `<JoinButton>` component should be the standard `<PrimaryButton>` and `<SecondaryButton>`, that get their actual theme from the community context (via `react-css-super-themr`'s `<ThemeProvider>`);
- The theme for the main component's button (which is rendered into the page and triggers the joining flow) should be optionally passable via the `<JoinButton>` props, because ad-hoc styling is necessary in some communities (for example, http://blockchain.topcoder-dev.com/: Join Now button in the Learn page uses the standard PrimaryButton style for this community, while the Join Now button in the Home page head banner has a bit different style.)
- To make two previous point for all communities, you might need to add new button themes (currently we have these in the codebase: http://community-west.topcoder-dev.com/examples/buttons/)

zjuasmn:
PR: https://github.com/topcoder-platform/community-app/pull/623
hash: 04d8c82

PR #623, opened 2017-11-29
#598, JoinCommunity button use the standard theming approach.

zjuasmn:


src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -35,30 +40,36 @@ export default function ConfirmModal({
...
-            <button
+          <div>
+            <PrimaryButton
               onClick={() => join(token, groupIds[0], userId)}
-              styleName="btnConfirm"
-            >Join</button>
-            <button
+              size="sm"
+              theme={theme.primaryBtn}
+              composeContextTheme={COMPOSE.SWAP}
```
birdofpreyru: No need to swap the theme here, and actually no need to pass custom theme into these internal modal buttons the way you do. The themes for these internal buttons should come from the community context; for that you should add `<ThemeProvider>` for each community, like here https://github.com/topcoder-platform/community-app/blob/develop/src/shared/routes/Communities/Veterans/Routes.jsx#L26
So if in that router of Veterans community you add
```
import primaryButtonTheme from 'components/buttons/outline/round/open-sans/blue-uppercase.scss';
```
and then, inside `<ThemeProvider>`'s `theme` object you add the entry
```
PrimaryButton: primaryButtonTheme,
```
Then all instances of `<PrimaryButton>` inside that `<ThemeProvider>` subtree will be styled with this context theme automatically. So, basically, we just need to add such `<ThemeProvider>` to each community, to specify default themes of primary and secondary buttons within each community (which are outline buttons for most of communities you can join yourself, and the default Topcoder buttons for the Veterans).

src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -94,4 +109,10 @@ ConfirmModal.propTypes = {
   resetJoinButton: PT.func.isRequired,
   token: PT.string,
   userId: PT.string,
+  theme: PT.shape({
+    primaryBtn: PT.shape().isRequired,
+    secondaryBtn: PT.shape().isRequired,
+  }).isRequired,
 };
+
+export default themr('ConfirmModal', { primaryBtn, secondaryBtn })(ConfirmModal);
```
birdofpreyru: I guess, when you specified `primaryBtn` and `secondaryBtn` themes here you though you pass them in as default styling, but actually because later you pass them into the buttons via their `theme` props, the button treat them as ad-hoc styles, thus breaking entire logic of context styling.

src/shared/components/tc-communities/JoinCommunity/index.jsx
```diff
@@ -40,7 +41,7 @@ export default function JoinCommunity({
     <div
       className={theme.container}
     >
-      <button
+      <GenericButton
```
birdofpreyru: Let's use default `<Button>` so that if we don't pass any particular theme via the component probs, the button will appear with the style of default buttons in that community.

src/shared/components/tc-communities/JoinCommunity/index.jsx
```diff
@@ -50,23 +51,24 @@ export default function JoinCommunity({
           }
           showJoinConfirmModal();
         }}
-        className={`${theme.link} ${state === STATE.JOINING ? style.joining : ''}`}
+        className={state === STATE.JOINING ? style.joining : ''}
+        theme={{ button: theme.link, disabled: style.disabled }}
```
birdofpreyru: I think, for the Join button itself, it will be better to pass entire theme as a dedication optional prop, say `joinButtonTheme`, and then here we pass into the button `theme={joinButtonTheme}`. It will be always used for ad-hoc correction of Join button style, so it is fine.

src/shared/components/buttons/outline/round/community/_mixin.scss
```diff
@@ -0,0 +1,49 @@
+@import "~styles/mixins";
```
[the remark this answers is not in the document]
zjuasmn: There is some difference between existed `outline/round/open-sans/COLOR-uppercase` and the newly added style (background, box-shadow...). But they are not used in any component,  may I replace it?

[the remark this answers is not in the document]
birdofpreyru: What is the difference? I'd say, just re-use existing theme, and msg me what is the difference.

zjuasmn:
Please check my updated PR.

src/shared/components/buttons/index.jsx
```diff
@@ -135,4 +137,10 @@ export const PrimaryButton =
 export const SecondaryButton =
   themr('SecondaryButton', secondaryButton)(GenericButton);
 
+export const PrimaryOutlineButton =
```
birdofpreyru: No, no need to add `PrimaryOutlineButton` and `SecondaryOutlineButton`, please remove them.
The whole idea behind the theming is that we already have all types of button we might need:
`PrimaryButton`, `SecondaryButton`, `Button` (default button), `DangerButton`, `GhostButton`, and we only use them everywhere, including in out modal; and then we use the context theming, i.e. if we render our join commuity modal within a subtree wrapped into
```
<ThemeProvider
  theme={{
    PrimaryButton: differentPrimaryButtonTheme,
    SeconardyButton: differentSecondaryButtonTheme,
  }}
/>
```
then primary and secondary buttons in our modal will automatically pick-up these styles.

src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -35,49 +36,42 @@ export default function ConfirmModal({
...
-            <button
-              onClick={resetJoinButton}
-              styleName="btnCancelLong"
-            >Cancel</button>
-          </span>
-        )}
-      </div>
+      { userId ? (
+        <div className={style.joinButtons}>
+          <PrimaryOutlineButton
```
birdofpreyru: should be `<PrimaryButton`

src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -35,49 +36,42 @@ export default function ConfirmModal({
...
+            onClick={() => join(token, groupIds[0], userId)}
+          >Join</PrimaryOutlineButton>
+          <SecondaryOutlineButton
```
birdofpreyru: should be `<SecondaryButton`

src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -35,49 +36,42 @@ export default function ConfirmModal({
...
+            onClick={resetJoinButton}
+          >Cancel</SecondaryOutlineButton>
+        </div>
+      ) : (
+        <div className={style.loginButtons}>
+          <PrimaryOutlineButton
```
birdofpreyru: `<PrimaryButton`

src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -35,49 +36,42 @@ export default function ConfirmModal({
...
+            onClick={() => {
+              const url = encodeURIComponent(
+                `${window.location.href}?join=${groupIds[0]}`,
+              );
+              window.location = `${config.URL.AUTH}/member?retUrl=${url}`;
+            }}
+          >Login</PrimaryOutlineButton>
+          <PrimaryOutlineButton
```
birdofpreyru: `<PrimaryButton`

src/shared/components/tc-communities/JoinCommunity/ConfirmModal/index.jsx
```diff
@@ -35,49 +36,42 @@ export default function ConfirmModal({
...
+                `${window.location.href}?join=${groupIds[0]}`,
+              );
+              url = encodeURIComponent(
+                `${config.URL.AUTH}/member?retUrl=${url}`,
+              );
+              url = encodeURIComponent(url);
+              window.location = `${config.URL.AUTH}/member/registration?retUrl=${url}`;
+            }}
+          >Register</PrimaryOutlineButton>
+          <SecondaryOutlineButton
```
birdofpreyru: `<SecondaryButton`

src/shared/components/tc-communities/JoinCommunity/index.jsx
```diff
@@ -50,23 +53,24 @@ export default function JoinCommunity({
...
-      </button>
+      </Button>
       { state === STATE.JOINED ? (
         <Modal onCancel={hideJoinButton}>
-          <h1 styleName="modalTitle">Congratulations!</h1>
-          <p styleName="modalMsg">You have joined the {communityName}!</p>
-          <button
+          <h1 className={style.modalTitle}>Congratulations!</h1>
+          <p className={style.modalMsg}>You have joined the {communityName}!</p>
+          <GenericButton
```
birdofpreyru: Should be `<Button`

zjuasmn:
@birdofpreyru Thank you for your detail explanation. Here is my updated PR.

Issue #478, opened 2017-09-20
[$200] - Switch of Challenge Details to API v3


birdofpreyru:
Now, a large part of challenge details is fetched from API v2. A new endpoint for challenge details has been added to API v3 (https://github.com/appirio-tech/ap-challenge-microservice/pull/192), thus we should move to use it as much as possible.

birdofpreyru:
### Current State & Expected Fix:
Right now, when we access a challenge details page, the actual details are pulled from:
1. `/v3/challenges?filter=id=CHALLENGE_ID`
2. `/v2/challenges/CHALLENGE_ID`
3. Some details might be pulled from other endpoints.

Actually, the new, and proper API v3 endpoint to pull challenge details from is the
`/v3/challenges/CHALLENGE_ID`, and we want to move to use it as much as possible
(keeping the old APIs only for the stuff we cannot get elsewhere). Note that this proper endpoint
may not return in its response some data present in (1).

So, here we expect the following:
- Add `getChallenge` method into [challenge service](https://github.com/topcoder-platform/community-app/blob/develop/src/shared/services/challenges.js); it should call the new API v3 endpoint, and, if the response misses some data served by (1), it should additionally call (1) and properly merge the responses (for that update and use the `normalizeChallenge` function inside the service: we should guarantee that the result returned from the challenge service always has a standard shape and content, so that any other code in Community App can rely on it, and does not need to make any duplicated checks of the challenge data content).
- Update the actions and reducer responsible for the challenge details page; so that they use the new service method, and it does not break anything in code.
- Clean-up any related obsolete code.

kondakovdmitry:
Could you answer some questions:
1. There seems to be a typo here: "if the response misses some data served by (1), it should additionally call (1) ". Did you mean data served by `/v3/challenges/CHALLENGE_ID`?
2. Should `getChallenge` fallback to extracting data from v2 API result in the case when v3 API call fails? (like it is done currently in [`extractChallengeDetail`](https://github.com/topcoder-platform/community-app/blob/develop/src/shared/containers/challenge-detail/index.jsx#L284))
3. I do not have access to [ap-challenge-microservice](https://github.com/appirio-tech/ap-challenge-microservice/pull/192), how can I get API v3 specs?

birdofpreyru:
@kondakovdmitry 
1. Right, I mean if `/v3/challenges/CHALLENGE_ID` misses something, then get it from (1), and mention in the code comments, what exactly is missing.
[...]

PR #625, opened 2017-11-30
#478 Switch of Challenge Details to API v3

kondakovdmitry:


birdofpreyru:
Hey Dmitry, very nice input! Accepted, but I'd like to ask for final fixes (and I've pushed a few commits into PR to merge with latest *develop* and to fix ESLint errors):
- `$ npm test` fails during unit testing:
  - `__tests__/shared/actions/challenge.js` fails because you have not added the new `getChalelngeDetails` function into the service mock for testing (https://github.com/topcoder-platform/community-app/blob/develop/src/shared/services/__mocks__/challenges.js). That mock is meant to be just a copy of the actual service, but instead of calling the actual API, it just gets a sample API response committed into the repo (right now it does only here https://github.com/topcoder-platform/community-app/blob/develop/src/shared/services/__mocks__/challenges.js#L124 for the V3 /challenges/ endpoint).
  - some other tests related to challenge also failing.
  please fix these.

kondakovdmitry:
@birdofpreyru Updated with fixed tests related to challenge. Could you review again, please?

Issue #602, opened 2017-11-22
[$150] - Use terms from v3 to display terms on a challenge


ajefts:
endpoint is ready in dev.  Terms are added to the challenge response:

http://api.topcoder-dev.com/v3/challenges/30028553

nauhil:
getChallengeTerms get from v3,  getCommunityTerms keeps use v2

right ?

kondakovdmitry:
`terms` in the v3 challenge response contains many duplicated items with the same `termsOfUseId` but with different `role`. Is it ok to show only terms with the `role` `Submitter` for a challenge?

birdofpreyru:
@kondakovdmitry yes, on challenge details page we need to show only `Submitter` terms, and only they are used to protect `Register` button.

kondakovdmitry:
The v3 challenge response does not contain `agreed` property which is required to check protection for `Register` button. How should we obtain whether user already agreed with the terms or not?

birdofpreyru:
@ajefts @skyhit can we add `agreed` property into terms included inside challenge details, if they were fetched with auth token?

kondakovdmitry:
PR: #632
hash: f40aa9f

This should work when `agreed` is added into terms inside challenge details.

kondakovdmitry:
@birdofpreyru Cannot test with DocuSign terms. When getting term details via v3 api, the following error response is always returned:
Response:
```json
{
    "id": "5daf7796:160bf4aaf4f:-7fcd",
    "result": {
        "success": true,
        "status": 500,
[...]

PR #632, opened 2017-12-03
#602 Use terms from v3 to display terms on a challenge

kondakovdmitry:


src/shared/containers/challenge-detail/index.jsx
```diff
@@ -299,7 +299,7 @@ const mapStateToProps = (state, props) => ({
   results: state.challenge.results,
   resultsLoadedForChallengeId: state.challenge.resultsLoadedForChallengeId,
   selectedTab: state.challenge.selectedTab || 'details',
-  terms: state.terms.terms,
+  terms: ((state.challenge.details || {}).terms || []).filter(t => t.role === 'Submitter'),
```
birdofpreyru: Good, but this is a wrong thing to do inside `mapStateToProps()`: this method is called now and then, and each time it is called, even if `state.challenge.details` is not changed, the `.filter()` method will create a new array each time; then ReactJS will receive a new array as `terms`, and will re-render the sub-tree where it is used. So, never ever create new objects inside `mapStateToProps()` to pass them into the return value.

The correct way is to make the same, somewhere in the underlying component (like `render()` method, etc.); cause the render method will be re-evaluated only if `mapStateToProps()` has returned a result different from the previous time.

kondakovdmitry: @birdofpreyru Thank you for the note. I'll fix it shortly.

birdofpreyru:
Hey @kondakovdmitry are you around?

First of all, sorry for the big delay with this PR. I have finally launched the new Bug Bash (https://www.topcoder.com/challenges/30061839, worth to register and make a dummy submission, if you wanna get paid; also you are sure welcome to work on other issues), thus just started looking into this again.

I've looked on all this with a fresh eye now, and realized that what you have done so far is only a part of the problem, we also need to update the terms service to work with v3 api (the swagger is here: https://github.com/topcoder-platform/tc-terms-service/blob/dev/swagger.yaml) and also update the mock of terms service to work with the updated service. I sure can rise the payment for all this extra work; so the question is whether you are available/willing to work on this in the nearest days?

birdofpreyru:
Yeah, I also pushed a new commit into this PR which resolves a merge conflict of your PR with the latest *develop*.

kondakovdmitry:
@birdofpreyru I updated the PR and also merged the latest develop into it. Not sure why all those commits appeared in this PR before my commits - thought it would show my changes only. Do I need to fix that somehow?

As for the update, it still has a problem with the 'agreed' property - it is still missed from the v3 challenge details response - that's why registering for a challenge would not work (it would always think that terms are not agreed).

Issue #607, opened 2017-11-27
[$150] - Basic Terms Page


ajefts:
We need a version of the terms page that we can link to in a non-modal form:

![image](https://user-images.githubusercontent.com/3812270/33274168-4da20d98-d35d-11e7-84a0-c23371f0bf86.png)

We used to have these links:  https://www.topcoder.com/challenges/terms/detail/21153/

I believe they are not working now that we have community-app serving /challenges.  I'd like to have community app provide functionality to display the terms under the same URLs.

birdofpreyru:
### Notes
- The page should support both the simple TC terms, and the DocuSign terms.
- In the case of the simple TC terms it should be styled with the standard TC text styling (https://github.com/topcoder-platform/community-app/blob/develop/src/styles/_mixins/_typography.scss - `@include tc-title` for the main title, `@include tc-heading-md` and `@inlcude tc-body-md` for the headers and paragraphs).

zjuasmn:
PR: https://github.com/topcoder-platform/community-app/pull/633
hash: 1f19c57

PR #633, opened 2017-12-03
#607, Basic Terms Page

zjuasmn:
basic terms: <http://local.topcoder-dev.com:3000/challenges/terms/detail/21193>
doc sign terms: <http://local.topcoder-dev.com:3000/challenges/terms/detail/21153>
not found terms: <http://local.topcoder-dev.com:3000/challenges/terms/detail/123>

### DIFF
--- a/src/shared/components/Terms/index.jsx
+++ b/src/shared/components/Terms/index.jsx
@@ -231,13 +231,7 @@ export default class Terms extends React.Component {
                           details={details}
                           docuSignUrl={docuSignUrl}
                           getDocuSignUrl={getDocuSignUrl}
-                          agreeTerm={agreeTerm}
-                          agreeingTerm={agreeingTerm}
-                          closeModal={onCancel}
                           loadingDocuSignUrl={loadingDocuSignUrl}
-                          viewOnly={viewOnly}
-                          agreed={selectedTerm.agreed}
-                          nextTerm={this.nextTerm}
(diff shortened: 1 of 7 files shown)


src/shared/components/Terms/index.jsx
```diff
@@ -231,13 +231,7 @@ export default class Terms extends React.Component {
                           details={details}
                           docuSignUrl={docuSignUrl}
                           getDocuSignUrl={getDocuSignUrl}
-                          agreeTerm={agreeTerm}
```
birdofpreyru: @zjuasmn I don't really understand, what for have you touched this file, and why have you removed these props? It breaks the terms agreement flow in a modal dialog. Well, I just add these back to the code myself before merging; but if there is some reason for this that I have missed, please comment.

zjuasmn: @birdofpreyru  Because My IDE told me that this props are redundant. I check the implement of `component/Terms/TermDetails` and found that the propTypes of TermDetails only requires `details`, `docuSignUrl`, `loadingDocuSignUrl` and `getDocuSignUrl` and the methods in it didn't contains code related to other props.

birdofpreyru: @zjuasmn oh, I've double-checked, and you are right. I did some mistake when testing originally.

Issue #611, opened 2017-11-28
[$250] - API/Network status icons


birdofpreyru:
From this ticket: https://github.com/topcoder-platform/community-app/issues/606 get the *icons.zip* file, which contains two icons we would like to show in the corner of screen when any API calls result in errors. For this ticket you'll create the PoC implementation of this:
- If icon(s) are shown, they should be shown in a small rectangular gray panel in the bottom-left corner of the page (fixed position). That panel should not be shown if no icons are displayed. When an icon is hovered, it should open a tooltip (same style as on the challenge listing page), explaining the meaning of the icon.
- Showing/hiding the icons should work similarly to the UI error modal message, i.e. we should add dedicated functions into https://github.com/topcoder-platform/community-app/blob/develop/src/shared/utils/errors.js which will allow to show that message from any place in the code.
- When firing an icon, we should pass in and keep in the memory the reason it is shown. For example, when some API call fails due to backend problem, we want to add the endpoint to a set of failed endpoints (probably with some extra information), so that if subsequent calls find that the API is up again, we can hide the icon. I don't have a clear view yet, how exactly it should work, but let's, at least, draft some mechanism for this.
- To showcase this in action, please update the [challenges service](https://github.com/topcoder-platform/community-app/blob/develop/src/shared/services/challenges.js) to use these icons in case of errors.

kondakovdmitry:
So, there are two icons - 'API Alert' icon and 'Connection Lost' icon. Is it correct that we need to show 'API alert' icon for API errors and 'Connection Lost' icon for network errors? Are these icons exclusive? I.e. when API error happened and then network error, we should show only 'Connection Lost' icon, right?

zjuasmn:
PR: https://github.com/topcoder-platform/community-app/pull/635
hash: 221b257

PR #635, opened 2017-12-04
#611. API/Network status icons

zjuasmn:
To trigger network error, in the challenges page, check offline mode.
image

To trigger API error, visit some not found page, like <http://local.topcoder-dev.com:3000/challenges/87654321>
image


src/shared/actions/status.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+import _ from 'lodash';
+import { createActions } from 'redux-actions';
+
+/**
+ * Export Actions for usage by Redux
+ */
+export default createActions({
+  STATUS: {
+    CLEAR_ALL_ERRORS_STATUS: _.noop(),
```
birdofpreyru: Let's rename these to be `..._ERROR_ICON...`? Like `CLEAR_ALL_ERROR_ICONS`, `SET_NETWORK_ERROR_ICON`?
The other thing, I would like more to have a single set of `SET_ERROR_ICON`/`CLEAR_ERROR_ICON` methods, which in addition to `url` and `message` would take `id` of the icon. We define a constant set of valid IDs in the actions module. This way, when we want to add more icons for different errors we won't have to add actions / reducers code, just to add a new ID into the set and properly load the icon asset itself.

And, I mean, we still keep `CLEAR_ALL_ERROR_ICONS`.

src/shared/containers/Status/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+  ApiErrors,
+  networkErrors,
+  clearApiErrorsStatus,
+  clearNetworkErrorsStatus }) {
+  if (_.isEmpty(ApiErrors) && _.isEmpty(networkErrors)) {
+    return null;
+  }
+  return (
+    <div styleName="container">
+      {!_.isEmpty(networkErrors) ?
```
birdofpreyru: Also here, I would love to have it more generic, so that it is easy to add more error icon types in future, without having copy/paste the same blocks of code here.

src/shared/containers/Status/style.scss
```diff
@@ -0,0 +1,66 @@
...
+  bottom: $base-unit;
+  opacity: 0.7;
+  transition: opacity 200ms ease;
+
+  &:hover {
+    opacity: 1;
+  }
+
+  svg {
+    margin-left: $base-unit;
```
birdofpreyru: well, in such cases you better make equal left and right margins (like half of the necessary spacing between the elements), then you won't have to set different margin for the first/last children of the container, and that will simplify the stylesheet and make easier its future modifications.

src/shared/reducers/status.js
```diff
@@ -0,0 +1,42 @@
...
+    [a.clearApiErrorsStatus]: (state, { payload }) => {
+      if (payload in state.ApiErrors) {
+        const newApiErrors = { ...state.ApiErrors };
+        delete newApiErrors[payload];
+        return { ...state, ApiErrors: newApiErrors };
+      }
+      return state;
+    },
+
+  }, initialState || { networkErrors: {}, ApiErrors: {} });
```
birdofpreyru: the structure of related state part (after you merge it with the errors state) should be something like this:
```
{
  ...
  errorIcons: {
    errorIconId: [{
      url,
      msg,
    }]
  }
}
```
I mean, the point is that you don't specify the actual types of the icon, keeping them dynamic. In fact, we can even permit to add unknow error types to show there (we'll just have to check that we don't have an icon for them, but we can add a generic error icon).

src/shared/services/challenges.js
```diff
@@ -252,7 +253,10 @@ export function normalizeMarathonMatch(challenge, username) {
  * @private
  */
 async function checkError(res) {
-  if (!res.ok) throw new Error(res.statusText);
+  if (!res.ok) {
```
birdofpreyru: Hmm... I think we should show API error only in case of 5xx error statuses. 4xx is either user doing something wrong (trying access something non-existing, or he is not authorized to access), either a bug in the frontend code. These we don't really want to show with icon, but with error modal messages (I mean, my reasoning about it is that server problems are kind of a permanent condition of the server, until somebody restarts it or somehow fixes the bug; while 4xx seems to be one-time-things, so if you just don't do the same bad thing again, you won't meet the error again).

zjuasmn:
@birdofpreyru PR updated.

birdofpreyru:
Hey @zjuasmn , sorry for the huge delay with reviewing and accepting the final version of your PR.
Please register to the new Bug Bash challenge (http://www.topcoder.com/challenge-details/30061839) and make a dummy submission there to get paid (and you are sure welcome to participate in fixing new issues as well :)

Issue #629, opened 2017-12-01
[$50] - More winners


birdofpreyru:
Check 
*RUX - 5Day - HLS Diabetes Story Visualization Design Concepts Challenge* in prod challenge listing. In the leaderboard it shows 3 winners, and another circle with `+1`. Actually, there are five winners, so should be `+2`. Also, on a wide screen there is enough space to show all.

PR #697, opened 2018-01-07
[$50] - More winners #629

nauhil:


birdofpreyru:
Oh, it was an easy fix, way easier than what I was expecting :)
Anyway, accepted.

Issue #445, opened 2017-09-08
Next Deadline msg on Challenge Details page for past challenges


birdofpreyru:
It has been fixed during Bug Bash, but it look like we missed some corner-cases (most probably, it is relevant to past F2F only, where data from API are a bit surprizing).

![screenshot](https://user-images.githubusercontent.com/20144632/30190620-a59bdb5c-943b-11e7-8ec2-1ffd288810f3.png)

birdofpreyru:
### Expected Fix:
It seems that this status msg still can be wrong in some corner cases. You should check all types of challenges in different tracks and states that you can find in dev and prod environment, and fix all discrepancies you can find.

For stalled challenges (in case they really don't have a winner) we should write there `The challenge is stalled`. If the current phase is late, and was not closed, we should properly report it as well, saying it is late (similar to how it is done in the challenge listing), etc.

nahidshahin:
Hi, can you please assign this ticket to me? I don't have the permission. Thanks

ThomasKranitsas:
@nahidshahin this ticket is not part of my bug bash.

PR #700, opened 2018-01-07
[$50] - Next Deadline msg on Challenge Details page for past challenges #445

imadx:


src/shared/components/Terms/index.jsx
```diff
@@ -160,7 +160,7 @@ export default class Terms extends React.Component {
           {
             !isLoadingTerms && (
               <div styleName="modal-content">
-                <div styleName="title">{terms.length > 1 ? 'Terms & Conditions of Use' : terms[0].title}</div>
+                <div styleName="title">{terms.length > 1 ? 'Challenge Prerequisites' : terms[0].title}</div>
```
birdofpreyru: This change seems to be related to another issue.

src/shared/components/Terms/styles.scss
```diff
@@ -44,6 +44,7 @@
   font-size: 28px;
   margin-bottom: 1.5 * $base-unit;
   text-align: center;
+  color: $tc-gray-80;
```
birdofpreyru: This as well related to another issue, please remove from this PR.

src/shared/components/challenge-detail/Header/index.jsx
```diff
@@ -187,12 +189,20 @@ export default function ChallengeHeader(props) {
   let nextDeadlineMsg;
   switch ((status || '').toLowerCase()) {
     case 'active':
-      nextDeadlineMsg = (
-        <div styleName="next-deadline">
-          Next Deadline: <span styleName="deadline-highlighted">
-            {nextDeadline || '-'}</span>
-        </div>
-      );
+      if (timeLeft === 0) {
```
birdofpreyru: This does not look correct to me. To check whether a challenge is stalled, we should look at the status field in the challenge object.

src/shared/components/Terms/index.jsx
imadx: My bad... I'll remove this.

src/shared/components/Terms/styles.scss
imadx: Will remove this as well.

src/shared/components/challenge-detail/Header/index.jsx
imadx: do we have a `stalled` status along with `active` and `completed` states?
where can I find a challenge in `stalled` competition to test?

birdofpreyru: @imadx I guess so, but I might be wrong. There are many stalled F2F in the dev challenge listing, and you can find out in the listing code, what is the actual conidition to distinguish stalled challenges.

imadx: @birdofpreyru The stalled competitions also shows their status to be `active`.
It doesn't seem to be the number of registrants either. Do you know anyone who knows what distinguishes stalled challenges?

birdofpreyru: Ok, I checked it myself, it is all here: https://github.com/topcoder-platform/community-app/blob/develop/src/shared/components/challenge-listing/ChallengeCard/Status/index.jsx
A challenge is stalled if it has no open phases beside registration phase.

imadx: Awesome! I'll fix this...

imadx: @birdofpreyru made the changes and checked with the challenge listing.
I hope the behavior is fixed now.

birdofpreyru:
I've checked against dev backend, there are many challenges that are late, but the message just shows 0 till the end of the phase, e.g.:

ishan-madhusanka:
Was that not what was needed? What was the required behavior?

birdofpreyru:
Have you read my *Expected Fix* post in the issue ticket?  For challenges being in a phase that is late to close we write `Late by ...` similar to how it is done in the challenge listing.

imadx:
@birdofpreyru Added the PR for this missing piece at https://github.com/topcoder-platform/community-app/pull/735

Issue #640, opened 2017-12-05
[$50] - Communities server-side service: no need to read from disk on each request


birdofpreyru:
We can read at start-up and serve from memory. A bit less convenient for local dev, but better performance for prod deployment.

The problem to fix is here: https://github.com/topcoder-platform/community-app/blob/develop/src/server/services/communities.js#L73

imadx:
added a PR to this issue: https://github.com/topcoder-platform/community-app/pull/701

imadx:
@birdofpreyru  updated the PR with the async file reads, according to your suggestion... https://github.com/topcoder-platform/community-app/pull/701

birdofpreyru:
@imadx Updated PR was functionally good enough to merge into the codebase, but I still want a bit more changes to make the code cleaner. I don't totally like your `groomMetadata` function, because you pass in `resolve`, which is confusing (to understand what is going on, somebody will have to track from which promise it originates); and also `groupsService`, which you cood just access as `this.private.groupsService` whould you do `groomMetadatada(..)` a private `this.private.groomMetadata(..)` method of the class instead. But, you don't even need to separate it into a dedicated function, if you do proper promise chaining. Now you have:

```
this.private.getMetadata = (communityId, knownGroups = {}) =>
  new Promise((resolve, reject) => {
    ...
    if (..) {
      ...
      groomMetadata(..);
    } else {
      ...
      fs.readFile(uri, 'utf8', (err, res) => {
        ...
        return groomMetadata(..);
      });
    }
  });
```

You better do:
```
this.private.getMetadata = (communityId, knownGroups = {}) => {
  const promise = new Promise((resolve, reject) => {
    ...
    if (...) {
      ...
      resolve(metadata); // These are non-groomed metadata
    } else {
      fs.readFile(uri, 'utf8', (err, res) => {
        if (err) {
          ...
          reject(...);
        }
        ...
        resolve(metadata); // These are non-groomed metadata
      });
    }
  });
  return promise.then((metadata) => {
    ... // Do the groomming here
    return metadata; // These are groomed metadata.
  });
};
```

Or you can keep the grooming code as a private method `this.private.groomMetadata(metadata, knownGroups)`; and then use it in the promise chaining doing in the end.

```
return promise.then(metadata => this.private.groomMetadata(metatada, knownGroups));
```
this is probably, the most clean option.

So, please do this, and will close the issue.

imadx:
[...]

PR #701, opened 2018-01-07
[$50] - Communities server-side service: no need to read from disk on each request #640

imadx:


src/server/services/communities.js
```diff
@@ -66,45 +68,53 @@ export default class Communities {
...
+        if (COMMUNITY_META_DATA[communityId]) {
+          metadata = COMMUNITY_META_DATA[communityId];
+        } else {
+          const uri = path.resolve(__dirname, '../tc-communities',
+            communityId, 'metadata.json');
 
-        /* Metadata itself, at the moment, are read from configuration files. */
-        fs.readFile(uri, 'utf8', (err, res) => {
-          if (err) {
+          metadata = fs.readFileSync(uri, 'utf8');
```
birdofpreyru: This is bad for two reasons:
1. Sync IO operation at server has impact on the performance for everybody. Sure, these reads happen only once for each community after each restart of the app, but still, let's follow the best practice and do it async.
2. They it was done before, if a non-existing community ID was passed in, the read operation failed, and the correct error message was generated. With your change, if `fs.readFileSync(..)` fails it will just crush the code (sure, we could wrap the call into `try {} catch(e) {}`, but see n1).

imadx: I'll fix this...

imadx:
used async `fs.ReadFile` to read files... but had to take out some of the implementation out of the original function to avoid code duplication.

PR #704, opened 2018-01-07
[$50] - More winners #629

imadx:


birdofpreyru:
I am sorry to dissapoint you, but another competitor resolved this issue first.
And the `TODO` comment was assuming a different thing anyway:
we have a standard, themeable Avatar component in our codebase: https://github.com/topcoder-platform/community-app/tree/develop/src/shared/components/Avatar
which should be used everywhere we need to show an avatar, thus the LeaderboardAvatar component (which does the same with a custom and less flexible code) should be just replaced at some point by the Avatar one, with any necessary updates of the code along the way. Though, as there no issues related to that, beside of the codebase quality, we can leave it as it until we really need some changes that touch the code in question that much, that it makes sense to make that replacement on the way.

imadx:
Cool... Thanks for the information.

Issue #647, opened 2017-12-08
[$100] - NDA differences between design and development


Dara-K:
As we talked earlier - check attached image to see the differences marked

![nda_differences](https://user-images.githubusercontent.com/25185079/33747658-dd50073e-dbcc-11e7-899d-de021a59dcae.png)

Dara-K:
Please let me know when you fixed this :)

Dara-K:
Also the buttons should stay at the bottom of the window like in the design - user can scroll. @birdofpreyru

imadx:
What can I do to reproduce this page?

birdofpreyru:
@imadx 
Option #1: Try to register to an NDA protected challenge in dev environment: https://www.topcoder-dev.com/challenges/30050950/
Option #2: Follow these instructions to mock-up terms service to always see the terms when you are trying to register for a challenge via the locally launched app: https://github.com/topcoder-platform/community-app/blob/develop/docs/mocking-terms.md

chongluo:
Can you please specify the behavior of the button? Now it is gone if agree clicked and next button does not show. Do you want the next button show after agree clicked?

birdofpreyru:
@chongluo I might be wrong now, but I believe the `Next` button shows up only when you look at a term that you have agreed with some time before, in this case it just moves to the next term of the challenge or community. If you look at a term not agreed before you see `Agree` button instead, and after agreement is done it automatically moves you to the next term, if any, or to the operation protected by the terms.

chongluo:
Another question about the blue and green tab. Can you please explain the logic when and which color show be shown? I might be wrong but I am assuming when a term agreed and tab and the line after it become blue, the terms the user reading but not click agree is green and the line after it is gray, and the rest of the tabs and lines are gray. Can you please confirm that?

birdofpreyru:
I guess, the current logic should be correct, most probably no need to modify that. I think, green circle is always the currently displayed term, no matter agreed or not, blue circles and lines are the agreed terms, and gray ones are the not yet agreed ones.

chongluo:
Should I change the title Terms & Conditions of Use or just keep it?

birdofpreyru:
@chongluo Let's expose it via props, and use the title from design for the challenge page, while keeping the current title as the default value (we use terms component to also show terms for sub-communities).

Dara-K:
[...]

PR #705, opened 2018-01-07
[$100] - NDA differences between design and development #647

jonadiazz:
Bugfix/ terms and conditions design. Also added line breaks for headers


birdofpreyru:
Sorry, but another competitor was first with a reasonably good fix.

PR #706, opened 2018-01-07
Fix minor typo

arun1595:
Fix minor typo firts -> first


birdofpreyru:
@arun1595 small typo, but still a good catch. I'll pay you $5 for this :)

PR #710, opened 2018-01-08
[$30] - Bug of Downloads section in Challenge details sidebar fixed #682

### LINKED ISSUE #682: [$30] - Bug of Downloads section in Challenge details sidebar
> Class C - https://www.topcoder.com/challenges/30060999/?type=design&tab=details <--- no downloads link available |  https://www.topcoder.com/challenges/30060999 <- Downloads now show....it's weird. Logged in. Chrome(latest). Win 10.

> birdofpreyru [22 hours ago] 
hm... while I was trying to check what is going on there, I found out that the downloads link has appeared again, at least for me.

> toxicpixel [7 hours ago] 
@birdofpreyru still not showing for me...".../?type=design&tab=details" is the only issue it seems. I remove that part of the address and PLOP there are the downloads.

> birdofpreyru [< 1 minute ago] 
Nope, I believe, if you reload the page, it will show you downloads on any url. We'll fix this bug later - good catch :slightly_smiling_face:

I guess the problem is this: downloadable documents are included into challenge details from api only when they are fetched with a valid auth token. So, if somebody goes directly to the page with expired token, the challenge specs are fetched as anonymous user, and don't include downloads. Once the page is loaded to the user, his auth token is automatically refreshed, but the specs are not updated, thus the downloads never appear.

### Expected Fix
If my guess is correct, then the challenge details container, in its `componentWillReceiveProps` should check the authorization segment of the state, and when `auth.user.userId` changes it should silently reload the challenge details with the new auth token.

andela-jomadoye:
This PR silently reloads the challenge details page with the new auth token when the userID changes. Fixes #682

### DIFF
--- a/src/shared/containers/challenge-detail/index.jsx
+++ b/src/shared/containers/challenge-detail/index.jsx
@@ -148,11 +148,16 @@ class ChallengeDetailPageContainer extends React.Component {
   componentWillReceiveProps(nextProps) {
     const userId = _.get(this, 'props.authTokens.user.userId');
     const nextUserId = _.get(nextProps, 'authTokens.user.userId');
+    const {
+      tokenV3,
+      authTokens,
+    } = this.props;
     if (userId !== nextUserId) {
       nextProps.getCommunitiesList(nextProps.authTokens);
     }
 
-    if (this.props.tokenV3 !== nextProps.tokenV3) {
+    if (tokenV3 !== nextProps.tokenV3
+      || authTokens.user.userId !== nextProps.authTokens.user.userId) {
       this.props.reloadChallengeDetails(nextProps.authTokens, this.props.challengeId);
     }
   }


src/shared/containers/challenge-detail/index.jsx
```diff
@@ -148,11 +148,16 @@ class ChallengeDetailPageContainer extends React.Component {
...
+      tokenV3,
+      authTokens,
+    } = this.props;
     if (userId !== nextUserId) {
       nextProps.getCommunitiesList(nextProps.authTokens);
     }
 
-    if (this.props.tokenV3 !== nextProps.tokenV3) {
+    if (tokenV3 !== nextProps.tokenV3
+      || authTokens.user.userId !== nextProps.authTokens.user.userId) {
```
birdofpreyru: It is not guaranteed that `authTokens.user` is always defined, that's why above in this function `_.get(..)` is used to fetch `userId` and `nextUserId` from `this.props` and `nextProps`. Anyway, if you compare auth tokens, then there is no need to append `|| TOKENS_ARE_DIFFERENT`: user ID is in fact encoded into the token, thus if the user IDs are different, then tokens are different as well.

There is one problem with your fix:
- When the app is opened, the v3 auth token is automatically refreshed about each 10 mins;
- Hence, your change will cause reloading of challenge details each 10 mins;
- Actually, we used to reload it this way before and it had a bad side-effect: if an user had opened challenge details and work on them, and then something happend with his Internet connection, or there is some instability of the TC backend, within 10 mins the challenge details reload will fail, and the already loaded challenge details will be dropped from the app.

So, to avoid that problem we should either (1) ensure that the details are reloaded only once, when we really need it to check the downloads; (2) to ensure that if reloading fails, but the challenge details have been already loaded before, then we don't drop them, and the code does not crush. We also should inform the user about problems with the reloading with the help of the API or network error icons (see the issue https://github.com/topcoder-platform/community-app/issues/611 and the related PR/commits already accepted into the codebase)

birdofpreyru:
@andela-jomadoye this PR has the issue I've explained in the comment in the code, thus rejected.

PR #721, opened 2018-01-08
Filter groups #58

### LINKED ISSUE #58: [$100] - Add groups filter to challenge listing page
On the challenges page, add a filter to allow the user to switch between the challenges that are in the current community (based on the group) or All challenges.  All challenges would be everything that is open (public) plus any challenges the user has access to via their groups.

We have some designs for this from challenges that chekspir ran.  We should use those designs.

arun1595:
Fixes #58 

The following info have been added to the sub community filter:

- Registration status of user, if authenticated
- Tooltip if user is not registered to the community
- Number of challenges in the community(after applying all active filters)

**Note**: I am having some hard time styling the tooltip, it would be awesome if you could help me with it
Relevant line: https://github.com/topcoder-platform/community-app/pull/721/files#diff-9c4b1e04b9bade397a95785108da3d71R97

Screenshot

image


birdofpreyru:
Not bad :) Accepted, but demands final fixes.

1. The way you treat `auth` all around is error prone. In prop checks you state it is non-required, and you default it to `null` all around, but in the places you access its fields, you don't check that `auth` is defined, which will crush the code if it is not. At the other hand, it is no need to assume it can be non-defined, as it always come from Redux state, and `state.auth` is always created by the reducer, so you can assume it is always defined, but then it should be clear from prop checks, and no default values of `auth` should be set.

2. You are missing green check marks next to registered challenges.

3. For some reason, the currently choosen sub-community does not show up in the dropdown selector when the dropdown is collapsed.

4. For **All** option let's add second-line text (same position and style as *registered / not registered* message, saying **Select to see all challenges**.

5. You are missing *Learn more* links for the not registered challenges (they should point to the home pages of these communities).

6. Not sure, what are your problems with the tooltip styling, but this is actually does not look that important to me as long as the tooltip is shown for all options (at the moment it shows up only for some options, and not for others), and contains the correct information. As long as you fix it, I can just correct the styling myself, or treat it as a separate small issue.

arun1595:
@birdofpreyru Thanks for the review. Let me address your comments.

arun1595:
@birdofpreyru 

> You are missing Learn more links for the not registered challenges (they should point to the home pages of these communities).

If we had Learn more links in the select items, clicking on the links will only choose the community and not direct the user to the linked page. Let me know if I am missing something here.

Btw, where can I find the urls for the community homepages? I tried using `communityId` to construct the urls but it doesn't work for all the communities.

birdofpreyru:
> If we had Learn more links in the select items, clicking on the links will only choose the community and not direct the user to the linked page. Let me know if I am missing something here.

Hmm... I am sure, it can be organized to work properly, so that the link is effectively above the option containing it, thus a click around the link selects the option, but a click on the link opens the community in a new page.

> Btw, where can I find the urls for the community homepages? I tried using communityId to construct the urls but it doesn't work for all the communities.

I don't remember exactly, what is the logic used by the community selector in the header nav of community pages, but we just should copy it.

arun1595:
@birdofpreyru I've ddressed all the points except 5. I've rendered 'Learn more' links to the communities user's not registered to. Not sure how to make sure a click on the link gets opened on a new page. Played with `z-index` a bit but couldn't get it to work.

Regarding 3, since I am rendering custom labels, I had to pass [valueRenderer](https://github.com/topcoder-platform/community-app/pull/721/files#diff-9c4b1e04b9bade397a95785108da3d71R184) prop to `react-select` to render the currently active community.

Regarding 6, I was previously assuming the tooltip should only be displayed if the visitor is not registered to the community. I've rendered the tooltip for all the community options except 'All' because it doesn't make sense to include 'You're registered or You're not registered...' for **All** option. I got the tooltip styling to work by resorting to `style` prop, not a clean solution though.

Let me know if I need to make any other changes. Thanks!

PR #722, opened 2018-01-08
Use api v3 for challenge register and unregister #479

### LINKED ISSUE #479: [$30] - Challenge register/unregister with API v3
Related functionality has been added to API v3 (https://github.com/appirio-tech/ap-challenge-microservice/pull/189). We should make necessary changes to use it instead API v2.

arun1595:
Fixes #479 

I made sure that unregistering from a challenge works but when I was trying to test registering a challenge, I keep getting a `Invalid URL http://localhost:3000/challenges/30051410/?type=develop` error. Not sure what's wrong with that.

### DIFF
--- a/src/shared/actions/challenge.js
+++ b/src/shared/actions/challenge.js
@@ -85,7 +85,7 @@ function getSubmissionsDone(challengeId, tokenV2) {
  * @return {Promise}
  */
 function registerDone(auth, challengeId) {
-  return getChallengesService(undefined, auth.tokenV2)
+  return getChallengesService(auth.tokenV3, auth.tokenV2)
     .register(challengeId)
     /* As a part of registration flow we silently update challenge details,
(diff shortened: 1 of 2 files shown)


src/shared/actions/challenge.js
```diff
@@ -85,7 +85,7 @@ function getSubmissionsDone(challengeId, tokenV2) {
  * @return {Promise}
  */
 function registerDone(auth, challengeId) {
-  return getChallengesService(undefined, auth.tokenV2)
+  return getChallengesService(auth.tokenV3, auth.tokenV2)
```
birdofpreyru: Here and below we use created service only to call the updated register / unregister methods, that do not depend on v2 auth token; l believe, we should remove auth.tokenV2 from the calls to `getChallengesService(..)`

arun1595:
@birdofpreyru Done. PTAL when you get a chance :smile:

PR #734, opened 2018-01-12
Add authorization middleware for auxiliary endpoints # 586

### LINKED ISSUE #586: [$100] - We should protected auxiliary endpoints in Community App's server by API key
Look through the server code module: https://github.com/topcoder-platform/community-app/blob/develop/src/server/server.js
There is a bunch of auxiliary endpoint, we should protect them by an API key.

For remote deployment the key should come from CircleCI environment variables. To track everything you need in code to make it happen, please search for all places where `FILESTACK_API_KEY` string is present (it is an unrelated api key, but it also comes from CircleCI for remote deployment).

arun1595:
Fixes #586


build.sh
```diff
@@ -31,6 +33,7 @@ fi
 TAG=$AWS_ACCOUNT_ID.dkr.ecr.$AWS_REGION.amazonaws.com/community-app:$CIRCLE_SHA1
 docker build -t $TAG \
   --build-arg NODE_ENV=$NODE_ENV \
+  --build-ard SERVER_API_KEY=$SERVER_API_KEY
```
birdofpreyru: Trailing `\` is missing

config/jest/default.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  collectCoverage: true,
+  // collectCoverage: true,
```
birdofpreyru: Not sure, why have you commented this? It is not related to the API key.

src/server/server.js
```diff
@@ -47,8 +47,17 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
 app.use(requestIp.mw());
 
+const checkAuthorizationHeader = (req, res, next) => {
+  const { authorization } = req.headers;
+  console.log('req', req.path, 'Authorization', authorization);
```
birdofpreyru: This debug output should be removed.

src/server/server.js
```diff
@@ -47,8 +47,17 @@ app.use(bodyParser.urlencoded({ extended: false }));
...
+  if (authorization !== process.env.SERVER_API_KEY) {
+    return res.status(500).send({ message: 'Invalid API key' });
```
birdofpreyru: Status - 403, no need for a specific message.

src/shared/services/money.js
```diff
@@ -60,7 +60,11 @@ function refresh() {
   }
   const url = isClientSide() ? '/community-app-assets/api/exchange-rates'
     : `${OE_API}/latest.json?app_id=${OE_TOKEN}`;
-  return fetch(url).then(res => res.json()).then((res) => {
+  return fetch(url, {
+    headers: {
+      authorization: process.env.SERVER_API_KEY
```
birdofpreyru:
```
Authorization: `ApiKey ${process.env.SERVER_API_KEY}`
```
To follow the standard look of the Authorization header, where the key or token is always prepended by the auth scheme name. Also requires corresponding update of the auth check method.

src/shared/utils/logger.js
```diff
@@ -47,6 +47,7 @@ if (isServerSide()) {
       }),
       headers: {
         'Content-Type': 'application/json',
+        'Authorization': process.env.SERVER_API_KEY
```
birdofpreyru: Node need to wrap `Authorization` into `'`. Also the value should match the format noted above.

arun1595:
@birdofpreyru Yeah, you're right. I wanted to make sure I was on the right track. I'll address your comments in a while. Thanks for the review :)

build.sh
arun1595: My bad :see_no_evil: 

Not sure how I missed it

config/jest/default.js
arun1595: This wasn't supposed to be in the PR, I'll clean this up.

arun1595:
@birdofpreyru `/api/tc-communities` endpoint already requires the v3 token to be passed via request headers. 

https://github.com/topcoder-platform/community-app/blob/develop/src/shared/services/communities.js#L25

Thus it causes a collision with the header authorization header check middleware that I am working on.

How should we go about this?

Thanks.

birdofpreyru:
@arun1595 no need to add API key auth to that one then.

__tests__/server/server.js
```diff
@@ -45,15 +45,23 @@ describe('Api test', () => {
   let server;
   beforeEach(() => {
     process.env.NODE_ENV = 'test';
+    process.env.SERVER_API_KEY = '79b2d5eb-c1fd-42c4-9391-6b2c9780d591';
```
birdofpreyru: I am not sure it is the best idea to set this for tests via env vars, better via `config/test.json`.

__tests__/server/server.js
```diff
@@ -45,15 +45,23 @@ describe('Api test', () => {
...
     jest.resetModules();
     server = require(MODULE).default;
   });
   test('post to /community-app-assets/api/logger', () => request(server).post('/community-app-assets/api/logger')
+    .set('Authorization', `ApiKey ${process.env.SERVER_API_KEY}`)
```
birdofpreyru: everywhere you read API key as `process.env.SERVER_API_KEY`, you should actually get it via Node config, i.e.
```
import config from 'utils/config'; // Cause Node config is server-side only, and our auxiliary utils/config modules hides the difference between client- and server-side configs.
...
.set('Authorization', `ApiKey ${config.SERVER_API_KEY}`)
```

src/server/server.js
```diff
@@ -47,8 +47,16 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
 app.use(requestIp.mw());
 
+const checkAuthorizationHeader = (req, res, next) => {
+  const { authorization } = req.headers;
+  if (authorization !== `ApiKey ${process.env.SERVER_API_KEY}`) {
```
birdofpreyru: Here it should be `Authorization` instead of `authorization`, and `config.SERVER_API_KEY` instead of `process.env.SERVER_API_KEY`, where `config` can be imported as `import config from 'config';`, cause this is server-side-only code.

src/shared/actions/rss.js
```diff
@@ -6,6 +6,8 @@ import fetch from 'isomorphic-fetch';
 import { toJson } from 'utils/xml2json';
 import { createActions } from 'redux-actions';
 
+const config = require('utils/config');
+
```
birdofpreyru: That's correct in this module

src/shared/components/tc-communities/NewsletterSignup/index.jsx
```diff
@@ -13,6 +13,8 @@ import { fireErrorMessage } from 'utils/errors';
 import qs from 'qs';
 import defaultStyle from './style.scss';
 
+const config = window.CONFIG;
+
```
birdofpreyru: This is not correct. This will work client-side only, but this code is under `/shared` folder, thus should work at the server-side as well (for server-side rendering), you should `import config from 'utils/config';`

src/shared/utils/logger.js
```diff
@@ -47,6 +47,7 @@ if (isServerSide()) {
       }),
       headers: {
         'Content-Type': 'application/json',
+        Authorization: `ApiKey ${process.env.SERVER_API_KEY}`,
```
birdofpreyru: This is server-side code only, should be `import config from 'config';` or `import config from 'utils/config';`, if you don't wanna think whether it is server- or client-side, and then `config.SERVER_API_KEY`: reading it directly from `SERVER_API_KEY` misses the case when the key is set via config, and not via the env var.

arun1595:
@birdofpreyru Have addressed your comments. Please take another look when you get a chance.

src/server/server.js
```diff
@@ -47,8 +48,15 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
 app.use(requestIp.mw());
 
+const checkAuthorizationHeader = (req, res, next) => {
+  if (req.headers.authorization !== `ApiKey ${config.SERVER_API_KEY}`) {
```
arun1595: The `req.headers` object received in the middleware has the `authorization` key and not the `Authorization` key.

PR #736, opened 2018-01-12
Mock morgan and logger in server tests #630

### LINKED ISSUE #630: [$100] - The unit test for server.js seems to randomly fail in CircleCI from time to time. Should be double checked why, and fixed.
This is the error message for the test:
image

The test is defined in this module: https://github.com/topcoder-platform/community-app/blob/develop/__tests__/server/server.js

arun1595:
Fixes #630


birdofpreyru:
@arun1595 I am not sure whether your changes solve the original problem of the occasional test failures, but while reviewing I believe I understood the problem:

- `logger.js` relies on `isServerSide(..)` to decide how the logger should be created, and it, in turn, relies on the `FRONT_END` env var. Some of the tests do change environment variables. It looks like sometimes jest runs the tests in such order and timing that at the moment of logger initialization `process.env.FRONT_END` turns out to be `true`, thus the frontend logger is created, which uses API calls, that fail in the test environment.

I am not sure now, how exactly `jest.genMockFromModule(..)` works, and whether it will securely prevent this from happening.

No need to mock morgan logger, I believe. Also not sure whether all changes you did in the test code are really necessary. In anyway your changes fail `npm test` due to lint errors.

So, please revise everything taking into account what I wrote above.

arun1595:
@birdofpreyru Thanks for the detailed review.


I mocked both `morgan` and `utils/logger` because those are external dependencies from `server.js` perspective. So, we don't really need to exercise them in the tests as we're only interested in testing the API.

>  It looks like sometimes jest runs the tests in such order and timing that at the moment of logger initialization `process.env.FRONT_END` turns out to be true, thus the frontend logger is created, which uses API calls, that fail in the test environment.

If we mock `logger`, we don't have to worry about the above scenario and trying to make sure `process.env.FRONT_END` is `undefined` before logger is created.

Am I making sense? Let me know your thoughts.

> Also not sure whether all changes you did in the test code are really necessary. In anyway your changes fail npm test due to lint errors.

Those are just cosmetic changes, I'll revert them back :)

birdofpreyru:
ok, then we are good, just ensure that your PR don't upset the linter.

arun1595:
@birdofpreyru I've cleaned up the mocks and the eslint is happy now :smile: 

Please take another look.

PR #746, opened 2018-01-15
Suppress warning mismatch between server and client render #603

### LINKED ISSUE #603: [$10] - LoadingIndicator triggers warnings about incorrect SSR
Because its states evolves right after the loading of the page, outside of ReactJS, so once ReactJS loads it sees that injected markup does not match with the browser's current content.

arun1595:
Fixes #603 

I was able to get rid of the warning due to the mismatch between client and server rendered `LoadingIndicator` by [this change](suppressHydrationWarning) but I am not entirely sure if it is okay as it will hide all hydration warnings thrown by react if we have any.

### DIFF
--- a/src/shared/utils/router/SplitRoute.jsx
+++ b/src/shared/utils/router/SplitRoute.jsx
@@ -117,6 +117,7 @@ export default class SplitRoute extends React.Component {
                   dangerouslySetInnerHTML={{
                     __html: window.SPLITS[`${TMP_CHUNK_PREFIX}/${chunkName}`],
                   }}
+                  suppressHydrationWarning
                 />
               );
               /* eslint-disable react/no-danger */


birdofpreyru:
Nope, just disabling all Hydration warnings is not good: we only want to disable them for the components that have a ligit reason to be ignored (at the moment only the loading indicators), in all other cases we need to keep the warnings to know that we should work on the issues.

birdofpreyru:
I am pretty sure, it is in fact caused by the loaders :) What happens is:
- Server renders markup of a page containing loading indicators and sends it to the client;
- Loading indicator animation code loads first and starts to animate (change params of the loading indicator circles) before the react bundle is loaded;
- When React bundle is loaded and initialized it uses `dangerouslySetInnerHTML` as a way to inject the target code into the prerendered markup;
- At this point React checks that the injected code does not match the current markup (cause that was already updated by the loading indicator animation) and prints you the warning

arun1595:
@birdofpreyru Yeah, it is in fact caused by the animation that changes the `opacity` and `radius` attributes of the server rendered `LoadingIndicator` but my point is passing `suppressHydrationWarning` to `LoadingIndicator`  won't have any effect as react's warning points to the topmost element.

birdofpreyru:
From React docs it looks to me that it is supposed to work for individal components, not only for the top-most ones; otherwise it does not make much sense.

arun1595:
Yeah, it looked like it and I did try passing `suppressHydrationWarning` prop to `LoadingIndicator` component, nothing changed. And then I tried passing it directly to the `circle` component but again, it didn't have any effect.

arun1595:
TBH, the react docs on `suppressHydrationWarning` confused me a bit and it doesn't make sense when the obvious change should suppress the warning.

PR #841, opened 2018-05-10
1st challenge attempt

meggamind:
There are a lot of things not included in this PR. 
I just wanted to push this as I worked a lot on this


birdofpreyru:
Hi @meggamind , which challenge is it related to? In any way, you should submit solution as Git patch via Topcoder submission system, and not create a PR directly on the repo.

PR #856, opened 2018-05-23
TopCoder challenge: Added Resources, Compete, Assets, Assets Details …

meggamind:
…Pages


kkartunov:
Low quality and after challenge end.

PR #857, opened 2018-05-25
Prevent multiple copies of React loaded

liuliquan:


birdofpreyru:
Not necessary. To prevent multiple copies of React or other dependencies, one should just ensure that exactly the same versions are required in the host code and in the libraries. It is so in the latest *new-topcoder-react-lib* branch.

Going forward, we just should enhance *adopt-dev-deps* script in *topcoder-react-utils* lib to take care about it.

PR #866, opened 2018-05-31
New settings profile

dedywahyudi:
@birdofpreyru 

This is the updated PR for New settings profile
FYI: The failed test on this PR because we need publish the topcoder-react-lib to the next version

Let me know if you find some issues. 

cc @cwdcwd @urwithat @ThomasKranitsas 

Thank you


src/shared/components/Settings/Profile/Skills/index.jsx
```diff
@@ -0,0 +1,158 @@
...
+import DevFallbackIcon from 'assets/images/profile/skills/id-develop.svg';
+import DesignFallbackIcon from 'assets/images/profile/skills/id-design.svg';
+import DataFallbackIcon from 'assets/images/profile/skills/id-data.svg';
+import { isomorphy } from 'topcoder-react-utils';
+
+import './styles.scss';
+
+let assets;
+if (isomorphy.isClientSide()) {
+  assets = require.context('assets/images/profile/skills', false, /svg/);
```
birdofpreyru: Why? I see no reason to import SVG only at clients-side.

src/shared/components/toastr.jsx
```diff
@@ -0,0 +1,99 @@
...
+            }
+          </span>
+        );
+      });
+  }
+}
+
+export default connect(
+  state => ({
+    toastr: state.toastr ? state.toastr : state.get('toastr'),
```
birdofpreyru: 1. Why connection to Redux in a component, and not in container?
2. Also I don't get the meaning of this line, why not just `state => ({ toastr: state.toastr })` ?

src/shared/utils/countries.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * Provides lookup data for countryies.
+ */
+import countries from './countries.json';
```
birdofpreyru: For such functionality we should rely on some standard external package, and not to implement it ourselves.

src/shared/utils/tc.js
```diff
@@ -250,4 +250,17 @@ export function processSRM(s) {
...
+ * @param {String} str1 One string
+ * @param {String} str2 Another string
+ * @returns {Boolean} true if two strings equal, false otherwise
+ */
+export function looseEqual(str1, str2) {
+  if (!str1 && !str2) {
+    return true;
+  }
+  return _.isEqual(str1, str2);
+}
```
birdofpreyru: 1. It is equivalent to `!str1 && !str2 || str1 === str2`
2. I believe, if we properly sets default props in our code, we'll never need such comparisions.

dedywahyudi:
@birdofpreyru Thanks for the detailed feedback! 
I'll get back to you

src/shared/components/Settings/EmailPreferences/index.jsx
```diff
@@ -0,0 +1,177 @@
...
+
+import LoadingIndicator from 'components/LoadingIndicator';
+import Styles from './styles.scss';
+
+const newsletters = [
+  {
+    id: 'TOPCODER_NL_GEN',
+    name: 'General Newsletter',
+    desc: 'News summary from all tracks and programs',
+  },
```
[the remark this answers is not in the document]
liuliquan: The existing Angular.js app is that way:

https://github.com/appirio-tech/topcoder-app/blob/dev/app/settings/email/email.controller.js#L20

src/shared/components/Settings/Profile/ExternalLinks/ExistingLinks.jsx
```diff
@@ -0,0 +1,259 @@
...
+  if (link.indexOf(startingUrl) === 0 || link.indexOf(httpsStartingUrl) === 0) {
+    result = link;
+  } else {
+    result = startingUrl + link;
+  }
+  return result;
+}
+
+// Open link function
+function openLink(e, link) {
```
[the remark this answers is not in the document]
liuliquan: It's used to open user's external link (like github profile page) in new tab

PR #1133, opened 2018-07-19
Services & subscriptions in settings page

ThomasKranitsas:


ThomasKranitsas:
This one should be good to merge but since CircleCI is down, let's wait :)

PR #1161, opened 2018-08-11
changing gender type prefer not to answer issue_1107 fixed

PrakashDurlabhji:


ThomasKranitsas:
@PrakashDurlabhji you also need to capitalize the other options.
`male` should become `Male` and `female` should become `Female`.

PrakashDurlabhji:
@ThomasKranitsas ok sorry forgot that , but i feel, it will capitalize the first letter due to css property, which is present.

ThomasKranitsas:
Yes but the data in the API call will have the options in lowercase which will throw an 400 BAD REQUEST ;)

PrakashDurlabhji:
@ThomasKranitsas just 1 min i will commit it.

PR #1196, opened 2018-08-15
Fixed issue 1011: copilot fulfillment should be consistent

nqviet:


src/shared/utils/memberStats.js
```diff
@@ -95,7 +95,8 @@ export function getHistory(stats, track, subTrack) {
 }
 
 
-const percentageFunc = n => `${(n * 100).toFixed(2)}%`;
+const percentageFunc = n => `${(n * 100).toFixed(0)}%`;
+const percentageFunc2 = n => `${n.toFixed(2)}%`; // fixed #1011
```
ThomasKranitsas: @nqviet please make the following changes:

1. The original `percentageFunc` had `toFixed(2)` while the new one you created has `toFixed(0)`. Any reason for that? Please reset that to the original.

2. Having enumerated methods with the same name (percentageFunc, percentageFunc2 etc) is not good as it'll be hard to maintain the code. Please find another name. Maybe something like `formatWithDecimal`?

nqviet: @ThomasKranitsas :
(1) => I have 2 repos, and I wrong picked my debug file when creating a pull request.
(2) => can the function body be:

const percentageFunc = n => {
&nbsp;  if (0 <= Math.abs(n) && Math.abs(n) <= 1) {
&nbsp;&nbsp;    return `${(n * 100).toFixed(2)}%`;
&nbsp;  }
  
&nbsp;  return `${n.toFixed(2)}%`;
};

and the function name still be `percentageFunc`, or `correctedPercentageFunc`?

nqviet: The above function body doesn't work if n = 0.5%, so I'll send a new update soon.

ThomasKranitsas: This would still be wrong because `percentageFunc(0.95)` will return `95%` while the value could actually be `0.95%`.
We need 2 different methods. One to calculate the percentage and one to format the number to `toFixed(2)`

PR #1207, opened 2018-08-16
convert to pwa

liuliquan:


sushilshinde:
I'll test the whole thing tomorrow.

liuliquan:
I think I made a wrong assumption that the service worker could be served from CDN with different domain. It looks like it has to be from the same origin domain.

E.g. When deployed in prod, it needs be from "https://www.topcoder.com/sw.js" or "https://community-app.topcoder.com/sw.js", but not from the cloudfront.net CDN

Then what would be the preferred path of the service worker?

liuliquan:
The reference: https://github.com/w3c/ServiceWorker/issues/940

ThomasKranitsas:
@sushilshinde ☝️

sushilshinde:
@liuliquan @ThomasKranitsas 

Path should be `https://www.topcoder.com/challenges`

liuliquan:
Done. The service worker path is at `/challenges/sw.js`.

And it needs be accessible from `https://www.topcoder.com/challenges/sw.js` and `https://community-app.topcoder.com/challenges/sw.js` when deployed in prod.

PR #1247, opened 2018-08-23
Fix side bar `here` link spacing

mahestro:
Challenge details page sidebar had the link ‘here’ without proper spacing in the paragraph


birdofpreyru:
@mahestro it is better to insert a new line between text and opening `<a>` tag, and put `{' '}` into that line, it will result in a regular whitespace, unlike the ubreakable ones used in your commit.

mahestro:
Thanks for the review @birdofpreyru. I applied the suggested approach.

Issue #1268, opened 2018-08-30
[$150] Submission - URL support in submission page


ajefts:
Provide support for entering a URL on the submission page instead of uploading a file.  We need to support 2 scenarios:

1.  The user provides the URL and we just store a reference to it (we don't attempt to download the file in the backend).
2.  The user provides a URL and we attempt to download the file on the backend.

Let's make sure we get the first one covered ASAP for topgear support.

ajefts:
@cwdcwd this one needs to be really high on the radar to get in there as soon as the core stuff is running ok.

cwdcwd:
queuing up

sushilshinde:
@ajefts @Dara-K 

Please have a look at the UI for this requirement

<img width="1503" alt="topcoder 2018-09-05 13-23-33" src="https://user-images.githubusercontent.com/683231/45079214-86dc2400-b10f-11e8-99cc-ec5ccd1b9a41.png">

Dara-K:
@sushilshinde I suggest we keep the Submission URL textfield outside the gray box as a regular field and we probably won't need the Submit button there, as we have one below, once the user checks that he agrees to the terms.

sushilshinde:
@birdofpreyru can we get this done 

**This is the requirement**

In the file stack component if the challenge group is top gear show only URL text field not other options for upload.

<img width="817" alt="111" src="https://user-images.githubusercontent.com/683231/45108226-ab5cee00-b159-11e8-9030-64f5f5625123.png">

suppermancool:
@birdofpreyru done
please check 
https://github.com/topcoder-platform/community-app/pull/1281

suppermancool:
@birdofpreyru done in https://github.com/topcoder-platform/community-app/pull/1281

suppermancool:
@birdofpreyru right now i check challenge group base on the url that enter challenge but look like it wrong, there is another option is check community id of challenge, but right now look like no challenge with wipro as community in dev enviroment. Can you create one topgear challenge to dev enviroment and give me the link to test, thanks

birdofpreyru:
@suppermancool not quite :)

ajefts:
@birdofpreyru @suppermancool @sushilshinde A few issues with this one.  Could we get these updates in asap?

1. Remove the restriction that it needs to be a zip file for URL submissions.
[...]

PR #1281, opened 2018-09-06
fix issue #1268 -> Set only url source for submission picker of topgear

suppermancool:
fix issue #1268 -> Set only url source for submission picker of topgear group

### DIFF
--- a/src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
+++ b/src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
@@ -100,7 +100,8 @@ class FilestackFilePicker extends React.Component {
       setFileName,
       setUploadProgress,
       uploadProgress,
-      challengeGroupName,
+      communitiesList,
+      groups,
     } = this.props;
 
     let pickupSources = [
@@ -111,8 +112,18 @@ class FilestackFilePicker extends React.Component {
       'github',
       'url',
     ];
-    if (challengeGroupName === 'topGear') {
-      pickupSources = ['url'];
+
+    if (communitiesList.data && communitiesList.data.length > 0) {
+      const topGearCommunity = _.find(communitiesList.data, { mainSubdomain: 'topgear' });
+      if (topGearCommunity) {
+        _.forOwn(groups, (value, key) => {
+          if (value && _.includes(topGearCommunity.groupIds, key)) {
+            pickupSources = ['url'];
+            return false;
+          }
+          return true;
+        });
+      }
     }
 
     return (
@@ -261,7 +272,6 @@ FilestackFilePicker.defaultProps = {
   error: '',
   fileName: '',
   uploadProgress: null,
-  challengeGroupName: '',
 };
 
 /**
@@ -271,6 +281,8 @@ FilestackFilePicker.propTypes = {
   error: PT.string,
   userId: PT.string.isRequired,
   challengeId: PT.number.isRequired,
+  communitiesList: PT.func.isRequired,
+  groups: PT.shape({}).isRequired,
   fileName: PT.string,
   fileExtensions: PT.arrayOf(PT.string).isRequired,
   title: PT.string.isRequired,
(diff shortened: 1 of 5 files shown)


suppermancool:
tested on this page:
http://local.topcoder-dev.com:3000/__community__/wipro/challenges/30052958/submit

suppermancool:
@birdofpreyru right now i check challenge group base on the url that enter challenge but look like it wrong, there is another option is check community id of challenge, but right now look like no challenge with wipro as community. Can you create one topgear challenge to dev enviroment and give me the link to test, thanks

ajefts:
@suppermancool You can try with this challenge:  https://www.topcoder-dev.com/challenges/30052982/

You can't use the URL.  I don't know if community ID is a safe way to check either (@birdofpreyru ?).

I believe the right way is to retrieve the groups returned from the challenge API and then check if those groups belong to topgear.  The group(s) belong to topgear if it is the parent topgear group (20000000) or any sub group of 20000000.

ajefts:
There are some endpoints to get parent groups and sub groups if it's needed:  https://github.com/appirio-tech/tc1-api-core/blob/master/tech.core/tech.core.service.identity/doc/swagger.yaml#L586

suppermancool:
@ajefts i can't load this challenge can you provide me your test account or any test account that can enter this challenge

suppermancool:
ah i can enter the challenge with another account, no need test account

suppermancool:
@ajefts @birdofpreyru
OK look like the group id work, the app already call the api that get list of group info (https://community-app.topcoder-dev.com/community-app-assets/api/tc-communities?groups%5B0%5D=1&groups). I think i can map the group id of the challenge with this list

sushilshinde:
@birdofpreyru please have a look, I will test once deployed

src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
```diff
@@ -100,8 +100,32 @@ class FilestackFilePicker extends React.Component {
...
+      'dropbox',
+      'onedrive',
+      'github',
+      'url',
+    ];
+
+    if (communitiesList.data && communitiesList.data.length > 0) {
+      const topGearCommunity = _.find(communitiesList.data, { mainSubdomain: 'topgear' });
+      if (topGearCommunity) {
+        _.forOwn(groups, (value, key) => {
```
birdofpreyru: 1. You must check that `communitiesList` is up-to-date (based on the timestamp in Redux store), and if it is not render the standard loading indicator instead of the filepicker. Otherwise, user will see all upload options if he opens filepicker before the list is loaded.

2. For optimal performance the logic should be reversed. First, check whether the challenge belongs to any group (most of challenges do not). If a challenge does not belong to any group, then we do not need to load / wait for `communitiesList` at all, we just render filepicker with all options on.

src/shared/containers/SubmissionPage.jsx
```diff
@@ -22,7 +27,18 @@ class SubmissionsPageContainer extends React.Component {
...
+    const {
+      auth,
+      communitiesList,
+      getCommunitiesList,
+    } = this.props;
+
+    if (!communitiesList.loadingUuid
+    && (Date.now() - communitiesList.timestamp > 10 * MIN)) {
+      getCommunitiesList(auth);
+    }
```
birdofpreyru: Here, we only need to load / update `communitiesList` if challenge belongs to any groups.

birdofpreyru:
@sushilshinde it has issue. In the corner case, when somebody with slow network rushes to open filepicker, he'll see all submit options until the request to load community data is fulfilled. It must be fixed. At the same time, I guess, many people are not that fast, and won't notice it. If you are in hurry, the current version can be rolled out, as it improves the current behavior in the correct direction, and once the further fixes are ready, we release them. Whatever you prefer.

birdofpreyru:
@sushilshinde @suppermancool yeah, PR closed cause I merged it to **develop**, but further fixes along my comments are needed.

src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
suppermancool: @birdofpreyru implemented in https://github.com/topcoder-platform/community-app/pull/1286

src/shared/containers/SubmissionPage.jsx
suppermancool: @birdofpreyru implemented in https://github.com/topcoder-platform/community-app/pull/1286

PR #1286, opened 2018-09-07
#1268 add ignore loading group and loading indicator 

suppermancool:
ignore loading group list if challenge don't belong to any group and adding loading indicator if the group haven't loaded yet
original issue: #1268

### DIFF
--- a/src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
+++ b/src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
@@ -129,7 +129,7 @@ class FilestackFilePicker extends React.Component {
           });
         }
       } else {
-        // show loading indicator if communitiesList isn't up-to-date and
+        // show loading indicator if communitiesList isn't up-to-date
         return (<LoadingIndicator />);
       }
     }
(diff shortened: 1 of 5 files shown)


src/shared/components/SubmissionPage/FilestackFilePicker/index.jsx
```diff
@@ -113,16 +114,23 @@ class FilestackFilePicker extends React.Component {
...
-          if (value && _.includes(topGearCommunity.groupIds, key)) {
-            pickupSources = ['url'];
-            return false;
-          }
-          return true;
-        });
+    // check if challenge belong to any group
+    if (!_.isEmpty(groups)) {
+      // check the group info and match with group list
+      if (communitiesList.data && communitiesList.data.length > 0) {
```
birdofpreyru: I checked reducer behavior, actually `tcCommunities.list.data`, from where `communitiesList.data` comes from, defaults to `[]`, thus the first part of the check makes no sense; and the second one is not correct, because a user can have no groups. The correct check will be to check that `tcCommunities.list.timestamp` is positive, which means that list has been fetched.

suppermancool: @birdofpreyru fixed tcCommunities.list.timestamp and styling of URL link button in the latest commit

PR #1590, opened 2018-09-27
issues-1305: Added Date picker and use at Birth Date

akmarif:
Added Date picker and use at Birth Date

Look of the date picker is now uniform across browsers. And at Edge in windows it's looks like this:
 
date_picker

### DIFF
--- a/src/shared/components/Settings/Profile/BasicInfo/index.jsx
+++ b/src/shared/components/Settings/Profile/BasicInfo/index.jsx
@@ -716,12 +716,14 @@ export default class BasicInfo extends ConsentComponent {
                   <label htmlFor="birthDate">
                     Birth Date
                   </label>
-                  <DatePicker
-                    numberOfMonths={1}
-                    date={newBasicInfo.birthDate}
-                    id="date-range-picker2"
-                    onDateChange={this.onUpdateDate}
-                  />
+                  <div styleName="date-picker-sm">
+                    <DatePicker
+                      numberOfMonths={1}
+                      date={newBasicInfo.birthDate}
+                      id="date-range-picker2"
+                      onDateChange={this.onUpdateDate}
+                    />
+                  </div>
                 </div>
                 <div styleName="field">
                   <label htmlFor="gender">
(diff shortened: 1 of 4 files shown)


ThomasKranitsas:
Here's my review:

1. The input field doesn't match with the other input fields.
2. On mobile, the popup is a bit off-screen.

akmarif:
working on it

akmarif:
Hi @ThomasKranitsas
> The input field doesn't match with the other input fields.
 Fixed at 0c7b55e

> On mobile, the popup is a bit off-screen.
I'm using same `z-index: 2` and other style like the date-range-picker at filter page. Can you please check that date-range-picker? Is that also has the same problem? can you please give me a screenshot?

ThomasKranitsas:
On mobile, there's no border

akmarif:
@ThomasKranitsas can I reproduce at mobile view of chrome? Can you please give me a screenshot?

ThomasKranitsas:
yeah, you can just resize your Chrome to reach the mobile breakpoint and you'll notice that the input doesn't match with the rest

akmarif:
@ThomasKranitsas I just did what you said, but don't see the problem. Here is the screenshot:

screen shot 2018-09-28 at 8 53 00 pm

ThomasKranitsas:
What do you mean you don't see the problem 🙂? Even in your screenshot, the birthdate input is way different than the other inputs (eg first name).

It shouldn't have that blueish background and it should have a border like the other fields.
Also, the font size and the padding is not correct.

ThomasKranitsas:
I'm talking about the actual input. Not the popup.

akmarif:
Ahh I see, fixing...

akmarif:
@ThomasKranitsas please check again, gave a fix at 01e3b58

ThomasKranitsas:
Don't worry about the `codacy/pr`. We don't use it.

ThomasKranitsas:
Better but still not perfect.
- [ ] The font is wrong
- [ ] The padding is wrong
- [ ] The border is cut-off in the corners.

akmarif:
@ThomasKranitsas 

> The font is wrong

Font fixed

>  The padding is wrong

Removed extra padding
 
> The border is cut-off in the corners.

border is better now

check at 548d357

PR #1591, opened 2018-09-27
tries to fix #1575

### LINKED ISSUE #1575: [$50] Submission Response Error Handling
**Description of bug**
Currently, when the submission API receives a `POST` of a submission the API checks that the user has the proper role applied to their JWT in order to POST. When the user does _not_ have the right role (or the roles are missing) the API returns a 403. The front end Community App is not handling this 403 correctly (and likely not handling any error codes) and is hiding the issue from the end user.

**To Reproduce**
1. POST submission to API with and invalid JWT
 

**Expected behavior**
Front end should gracefully handle and report error to end user

**Additional context**
This is most often seen with Topgear SSO users that have a botched/missing role set

sumitdaga:


src/shared/reducers/page/submission/index.js
```diff
@@ -39,10 +39,10 @@ function onSubmitDone(state, { error, payload }) {
   /* TODO: I am not sure, whether this code is just wrong, or does it handle
    * only specific errors, returned from API for design submissions? I am
    * adding a more generic failure handling code just above. */
-  if (payload.error) {
+  if (payload.result && !payload.result.success) {
```
cwdcwd: @sumitdaga I don't think this will do it. The v5 payload does not put errors in the response like that. We need to handle the issue on the HTTP Status codes

PR #1604, opened 2018-09-28
issues-1326: changed drop down text style

akmarif:
Changed drop down text style at given page and all sub pages

### DIFF
--- a/src/shared/components/Settings/style.scss
+++ b/src/shared/components/Settings/style.scss
@@ -157,7 +157,11 @@ $checkbox-bg-selected: $tc-dark-blue;
     .Select-placeholder,
     .Select-value-label,
     .Select-menu-outer {
-      @include merriweather-sans-regular;
+      font: {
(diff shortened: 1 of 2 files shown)


src/shared/components/Settings/style.scss
```diff
@@ -157,7 +157,11 @@ $checkbox-bg-selected: $tc-dark-blue;
     .Select-placeholder,
     .Select-value-label,
     .Select-menu-outer {
-      @include merriweather-sans-regular;
+      font: {
+        size: 15px;
+        family: "Roboto", Helvetica, Arial, sans-serif;
```
ThomasKranitsas: You should use the existing mixins for fonts instead of hardcoding the values.

akmarif: There was to existing mixin so I have to add a new on at fcb0bcd376cbedd377fac5540860edcb5cc47203

Issue #1436, opened 2018-09-14
[$10] Add external link textbox is set to sentence case on iPad


PereViki:
**Describe the bug**
When I start to type into the external link textbox, the input starts with an uppercase letter, which is not adequate for urls.

**To Reproduce**
1. Go to 'https://www.topcoder-dev.com/settings/account'
2. Click on 'Linked Account'
3. Start to fill the 'External Link' textbox

**Expected behavior**
The input should be lowercase.

**Screenshots**
![img_0526](https://user-images.githubusercontent.com/35376779/45571937-3039bd00-b868-11e8-97bd-6bc1d5f4e02e.PNG)

**Tablet:**
 - Device: iPad Mini 2
 - OS: iOS 11
 - Browser: Chrome
 - Version: 69.0.3497.91

@ThomasKranitsas

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30072042 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30072042 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
@topcoder-platform/topcodercompetitors this is open for pickup

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30072042 has been updated - it has been assigned to diazz.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

suppermancool:
done in https://github.com/topcoder-platform/community-app/pull/1629

ThomasKranitsas:
Payment task has been updated: https://software.topcoder.com/review/actions/ViewProjectDetails?pid=30072042<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

PR #1629, opened 2018-09-29
fix issue #1436

suppermancool:
fix issue #1436


src/shared/components/Settings/Account/LinkedAccount/AddWebLink.jsx
```diff
@@ -105,6 +106,7 @@ export default class AddWebLink extends React.Component {
                   </label>
                   <div styleName={webLinkValid ? 'validation-bar url' : 'validation-bar url error-bar'}>
                     <input
+                      autocapitalize="off"
```
ThomasKranitsas: I think the correct attribute for React is `autoCapitalize`

suppermancool: @ThomasKranitsas fixed in latest commit

PR #1686, opened 2018-10-11
issues:1685: Changed style of New password field

nahidshahin:
Using the same style as `Password`. Now its look the same:

screen shot 2018-10-11 at 9 29 13 pm

### DIFF
--- a/src/shared/components/Settings/Account/MyAccount/index.jsx
+++ b/src/shared/components/Settings/Account/MyAccount/index.jsx
@@ -585,7 +585,7 @@ export default class MyAccount extends React.Component {
                           </div>
                         </div>
                         <div styleName="row">
-                          <div styleName="field col-1 password">
+                          <div styleName="field col-1">
                             <label htmlFor="new-password">
                                 New password
                             </label>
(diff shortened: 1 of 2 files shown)


src/shared/components/Settings/Account/MyAccount/index.jsx
```diff
@@ -585,7 +585,7 @@ export default class MyAccount extends React.Component {
                           </div>
                         </div>
                         <div styleName="row">
-                          <div styleName="field col-1 password">
```
ThomasKranitsas: I remember we've added this class to fix the issue where the label was out of place when the tooltip was shown.

Can you please take a look at that?

nahidshahin: @ThomasKranitsas sure, looking into it.

nahidshahin: Hi @ThomasKranitsas, you are right. I missed the purpose the `password` class. I bring it back and removed border & shadow. Can you please check once more. Thanks

PR #1707, opened 2018-10-18
fix the link

nahidshahin:
As we are sending absolute url now, so we don't need `<Link>` component any more.


nahidshahin:
done!

PR #1715, opened 2018-10-20
Fix issue 1699

jomadoye:


ThomasKranitsas:
The fix is not quite correct. What we need instead is to be able to deep-link to an FAQ/Extra element and not the whole section.

We also need to be able to auto expand that section.

ThomasKranitsas:
Closing this as the changes are invalid

Issue #1738, opened 2018-11-05
[$100]Fix listing issues : Stalled issues and bucket issues (It's urgent)


sushilshinde:
For the branch `mm-develop-merge`
1. Fix Stalled issues, which are not there on the production

<img width="1650" alt="mm-deveop-merge" src="https://user-images.githubusercontent.com/683231/47982785-029a2300-e0f7-11e8-8f49-0e96d95a0986.png">

2. These stalled items are missing from `Open for registration` bucket they should be under this bucket.

3. My Challenges bucket isn't displayed - all the challenges which I'm registered for should be under this bucket

Compare results with topcoder.com/challenges(v3 API) for reference.

Run in production mode(v4 API), there is data model difference in v3 and v4 API.

Instruction to run an app in production mode

```
rm -rf node_modules
npm i
npm run build
npm start
```

Check **Deployment and Execution** section in the doc for details 
https://github.com/topcoder-platform/community-app/blob/develop/README.md

Create PR for changes, I'll reopen for pickup if the PR not submitted within 8 hours

**Repos :** 

https://github.com/topcoder-platform/community-app/tree/mm-develop-merge
https://github.com/topcoder-platform/topcoder-react-lib#new-develop-mm

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30073811 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30073811 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30073811 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

nahidshahin:
Can I work on this, please?

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30073811 has been updated - it has been assigned to rashmi73.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

sushilshinde:
@topcoder-platform/topcodercompetitors this one is open for pickup

rashmi73:
@nahidshahin i took this once @sushilshinde added open for pickup tag.

ThomasKranitsas:
[...]

PR #1746, opened 2018-11-08
Fix for issue #1738

gurmeetb:
Fix for issue #1738


sushilshinde:
@gurmeetb point 2 and 3 are missing from the submission

gurmeetb:
@sushilshinde sorry I missed them completely. Focus moved more on why `Stalled` ? let me see if I can get them immediately.

gurmeetb:
@sushilshinde let's open it for pickup, this will be again time consuming effort to find the fix for other 2. The fix is needed in `src/components/challenge-listing/Listing`. 

Kindly open and see if else who can pick-up and complete it. if no one is doing it, I will take a look after tomm afternoon IST. Currently I got some other deliverables right now so don't want to block you.  Sorry about the same.

Issue #1816, opened 2018-12-12
[$60]Challenge List Filter Issue


m-yadav:
**Describe the bug**
When the user filter challenges by tracks and after reload the page, the filters by track got lost, even though user can see them in the URL and this has been working before (at least a month ago) as reported by the user

**To Reproduce**
1. Open  https://www.topcoder.com/challenges
2. Turn off tracks Design and Data Science so only Development track left. You could see that list of challenges now has only development challenges as expected and URL become like this: https://www.topcoder.com/challenges?filter[tracks][develop]=true. So this URL suppose to show only development challenges.
3. Now either press Reload in the browser or navigate directly by link https://www.topcoder.com/challenges?filter[tracks][develop]=true
All the challenges will be shown and switches for all three tracks will be enabled.

See demo video https://monosnap.com/file/SxUxy58XczqQvyEHZGlOue83jgwEee

**Expected behavior**
only Development track switch is enabled and only development challenges are listed.

**Screenshots**
NA

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [Chrome]
 - Version [70.0.3538.110]

**Smartphone (please complete the following information):**
NA

**Additional context**
Note, that if user filter only challenges "open for registration" the URL will become like this https://www.topcoder.com/challenges?filter[tracks][develop]=true&bucket=openForRegistration and after page reload filter "open for registration" will persist as expected. So the only filter by tracks got lost as far as the user sees.

sushilshinde:
PR on
https://github.com/topcoder-platform/community-app/tree/hot-fixes

Prod mode setup locally
If you want run app in prod mode locally refer README

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30077021 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30077021 has been updated - it has been assigned to gurmeetb.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30077021 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

sushilshinde:
try this, I deployed your code here 

https://beta-community-app.topcoder.com/challenges?filter[tracks][develop]=true

[...]

PR #1832, opened 2018-12-18
Issue #1816

gurmeetb:
fix no filterState being passed to ChallengeFilters


gurmeetb:
@sushilshinde this should fix it. Once we test, I will revert the logs. Sorry about iterations but locally we don't see this issue.

PR #1835, opened 2018-12-18
issue_1786 fixed

Gauravseta:


sushilshinde:
@Gauravseta can you remove lint errors

PR #1974, opened 2019-02-14
Update sw.js

PrakashDurlabhji:


sushilshinde:
Build failed for this but commenting this code isn't the solution. 

Also earlier PR dint work as reload problem continues

PR #2233, opened 2019-03-18
Update countries in profile settings

abdih:
Transition away from using hard-coded dropdown values for list of countries on a user's Profile.


ThomasKranitsas:
Why did you create a PR? The challenge is still in review. Only the winner has to create a PR.

PR #2236, opened 2019-03-20
Fix issue with accounts that have only numbers in their handle

ThomasKranitsas:
This is a fix for a blocker issue for many TopGear members.

It fixes the issue that some pages (eg submission page) are not accessible if the handle contains just numbers.


ThomasKranitsas:
@sushilshinde it would be great to push this to prod ASAP as this blocks TopGear members from submitting in challenges

PR #2240, opened 2019-03-21
Added isSaving state to disabled prop of save btn

frankeros:


src/shared/components/Settings/Profile/BasicInfo/index.jsx
```diff
@@ -911,7 +916,7 @@ export default class BasicInfo extends ConsentComponent {
         <div styleName="button-save">
           <PrimaryButton
             styleName="white-label"
-            disabled={false}
+            disabled={this.isSaving}
```
ThomasKranitsas: It should be `this.state.isSaving`.

Your code is not working currently as `this.isSaving` is not defined

frankeros: I'm sorry, change the code and forget that part.
the new commit is already

PR #2247, opened 2019-03-21
issue 2105 fixed

rashmi73:


ThomasKranitsas:
@rashmi73 the fix must be done for all pages (eg devices, service providers etc), not just for the skills.

Please update your PR to fix the other pages as well.

Thank you

PR #2255, opened 2019-03-22
Issue 2083 fix

rashmi73:


ThomasKranitsas:
@rashmi73 the fix must be done for all pages not just the Devices

PR #2266, opened 2019-03-23
Fix for issues-2179

nahidshahin:
Showing Country Rank `-` if servers returns 0

### DIFF
--- a/src/shared/utils/memberStats.js
+++ b/src/shared/utils/memberStats.js
@@ -303,7 +303,7 @@ export function getDetails(stats, track, subTrack) {
           {
             key: 'rank.countryRank',
             label: 'Country Rank',
-            render: defaultRenderFunc,
+            render: x => (x === 0 ? '-' : x),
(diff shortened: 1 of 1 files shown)


nahidshahin:
@sexta13 
> I would do a different check. Something like x ? x : '-'.

I guess we don't need ternary operator at all, this one should work `x || '-'`

sexta13:
> @sexta13
> 
> > I would do a different check. Something like x ? x : '-'.
> 
> I guess we don't need ternary operator at all, this one should work `x || '-'`

True :) Well done

ThomasKranitsas:
@sexta13 how come you reviewed this one? 🙂

sexta13:
@thomaskranitsas.. I just commented because I was going to pick this one... But once I saw the issue was coming from the API..and in the issue there was no "expected result"... I came to see what was the implementation :) I just gave my 5cents of opinion eheh

PR #2291, opened 2019-04-04
Password change f2f

sumitdaga:


ThomasKranitsas:
@sushilshinde this is the fix for the updated flow for changing the password.

cc @ajefts @cwdcwd

PR #2370, opened 2019-05-24
add verified badge to skills

sumitdaga:


luizrrodrigues:
This PR is just to **Frontend**, **Backend API** already returning **CHALLENGE** and **USER_ENTERED** in `source` of `skills` data.

QA Question:
- Backend API already add **CHALLENGE** source to skill when member win a Challenge?

Looks like not working, check this profile below, you can see **QA** in skill WITHOUT **Verified** mark:
https://beta-community-app.topcoder.com/members/sdgun   
image

But check this Challenge (member on this Challenge), have **QA** in tech:
https://beta-community-app.topcoder.com/challenges/30069360
image

So, if member win Challenge with **QA** tag, why API doesn't return **CHALLENGE** to QA skill?

Issue #2391, opened 2019-05-31
[$100] Beta-Community-app: Design, Development, DataScience filter combinations are not working correctly


nithyaasworld:
**To Reproduce**
1. Go to https://beta-community-app.topcoder.com/challenges
2. Try various combinations of Design, Dev and Data Science filters

**Actual behavior**
1. When Design and Data Science filter is selected, it displays only Design Challenges
2. When Design and Dev alone are selected, it displays all challenges including Data Science

nithyaasworld:
Tagging @sushilshinde

ThomasKranitsas:
@topcoder-platform/topcodercompetitors this is open for pickup.

PR on `server_side_filtering` branch.

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30092402 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30092402 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

mhikeo:
@ThomasKranitsas Kindly assign this to me. Thanks.

ThomasKranitsas:
@mhikeo I'll ask an admin to add you to the org. You can start working on this and I'll assign this later to you.

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30092402 has been updated - it has been assigned to mhykol.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Payment task has been updated: https://software.topcoder.com/review/actions/ViewProjectDetails?pid=30092402<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

PR #2409, opened 2019-06-05
Fix for Issue #2391

mhikeo:


ThomasKranitsas:
Other combinations seem to work but not the above one.

mhikeo:
Checking ...

mhikeo:
@ThomasKranitsas Added the fix. Kindly check.

Issue #2405, opened 2019-06-04
$20 | New Nav: User Menu popup is overlapping with the Sub Community


nithyaasworld:
**To Reproduce**
1. Go to https://test-community-app.topcoder-dev.com/challenges
2. Login
3. Click on the user name present at the right side of the new nav
4. Check the User popup 

**Actual behavior**
Sub  Community is overlapping with the User Popup

**Screenshots**
![image](https://user-images.githubusercontent.com/5114190/58901903-b6ac6c80-871f-11e9-9fbd-be58bac21b17.png)

iversonLv:
Can I take this?

PrakashDurlabhji:
even I wanted to take it, confused whether it is actually open?
@luizrrodrigues

luizrrodrigues:
@iversonLv sure.

@PrakashDurlabhji issues with nav_OpenForPickup

iversonLv:
@luizrrodrigues I don't have permission set assignee and change label.
Please help with changing

Thanks.

luizrrodrigues:
Assigned to @iversonLv

luizrrodrigues:
I'll do this manually, maybe I can't assigned in right way too, but I'll
add a comment "Assigned to... "

luizrrodrigues:
@iversonLv Any status on this?

PrakashDurlabhji:
@luizrrodrigues can i give it a try if iverson is busy.

luizrrodrigues:
@PrakashDurlabhji If @iversonLv no reply in 2h from, yes.

iversonLv:
@luizrrodrigues I have raise the PR? 
https://github.com/topcoder-platform/community-app/pull/2589

iversonLv:
4 Days ago.
Maybe I need comment here the PR number here, sorry for inconvenient.

luizrrodrigues:
@iversonLv Ah sorry, I don't see this. Thanks :)

PR #2589, opened 2019-06-08
issue_2405 New Nav: User Menu popup is overlapping with the Sub Commu…

iversonLv:
…nity

PR for #2405


iversonLv:
@luizrrodrigues here

PR #2621, opened 2019-06-13
issue 2542 fix

rashmi73:


rashmi73:
@luizrrodrigues is "match" need to be used of react-router-dom?
 
but then it would be huge change.

Already tried with "withRouter" HOC but all fields are blank in it, because it takes relative values and Header is not wrapped by any Route Tag.

luizrrodrigues:
@rashmi73 Wait, need apply both PR to fix that right?

luizrrodrigues:
Deployed both, but not working here.

Working in your local, can you share a video?

PR #2624, opened 2019-06-14
Fix for issues#2519

nahidshahin:
Pass openMore shared variable between TopNav and LoginNav


luizrrodrigues:
https://github.com/topcoder-platform/navigation-component/pull/7

Issue #2638, opened 2019-06-18
New Nav: Page should not reload when navigating between menus


nithyaasworld:
In the New nav, when navigating between different menus, currently the total page is getting reloaded including the nav bar.

Expected: When navigating between different menus, The top navigation bar should remain static and only the content area should get refreshed with corresponding data.

r0hit-gupta:
May I assign this to myself?

luizrrodrigues:
@r0hit-gupta yes, thanks.

luizrrodrigues:
Assigned to @r0hit-gupta

r0hit-gupta:
It says ```Unauthorized``` whenever I click some link present on the header. I already logged in using the test credentials - ```dan_developer```. How do I resolve this?

luizrrodrigues:
@r0hit-gupta Need set ENV VARS, please check here:
https://apps.topcoder.com/forums/?module=Thread&threadID=938763&start=0&mc=1

If you not able to access forum, please register in this challenge:
https://www.topcoder.com/challenges/30092991/?type=develop

r0hit-gupta:
Thanks @luizrrodrigues. The app loads fine now. However, upon closer inspection I believe the actual bug lies in the ```navigation-component```. It uses ```<a>``` tags instead of ```<Link>``` tags which will always trigger a complete page reload. Should I make a PR on that repository?

luizrrodrigues:
Yes, please send PR to navigation-component if needed. Thanks

r0hit-gupta:
@luizrrodrigues sure. By the way it's late here. Do you mind if I do it in the morning?

luizrrodrigues:
We need move this ASAP

Em qua, 19 de jun de 2019 15:59, Rohit Gupta <notifications@github.com>
escreveu:

r0hit-gupta:
Unable to resolve the issue. Opening up for others.

luizrrodrigues:
@r0hit-gupta Okay, no problem. Thanks try that.

gets0ul:
Can I work on this?  
Thanks.

luizrrodrigues:
@gets0ul Yes, thanks.

gets0ul:
@luizrrodrigues It's been over 12 hours, you can unassign me from this. Thanks.

luizrrodrigues:
@gets0ul Okay, thanks for trying.

lucyji1000:
[...]

PR #2652, opened 2019-06-21
Fix for issue #2638

lucyji1000:
Fix for [this issue](https://github.com/topcoder-platform/community-app/issues/2638).

Based on my understanding of the `Link` component in `topcoder-react-utils` from the docs [here](https://github.com/topcoder-platform/topcoder-react-utils/blob/master/docs/link-and-navlink.md), menu links in the `TopcoderHeader` component were setting the `enforceA` prop of the `Link` component to `true` and forcing a reload of the entire page.

### DIFF
--- a/__tests__/shared/components/TopcoderHeader/__snapshots__/index.jsx.snap
+++ b/__tests__/shared/components/TopcoderHeader/__snapshots__/index.jsx.snap
@@ -39,7 +39,7 @@ exports[`Default render 1`] = `
         onTouchStart={[Function]}
       >
         <Link
-          to="https://www.topcoder-dev.com/community/learn"
+          to="/community/learn"
         >
           Tracks
         </Link>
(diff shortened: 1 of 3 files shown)


luizrrodrigues:
@lucyji1000 thanks, I'll review shortly

ljipear:
@luizrrodrigues I based this PR off the `bug-bash` branch as specified in the challenge instructions, but I'm seeing some differences with the `feature-new-nav` branch like additional menu items in the `tc-communities/Header` component that may also be affected by the page reload issue -- should I have branched off the `feature-new-nav` branch instead?

luizrrodrigues:
@ljipear Need to use `feature-new-nav` branch not `bug-bash`

lucyji1000:
Ah I was working off the `bug-bash` branch, which did not have the new nav header. I switched to the `feature-new-nav` branch, and the issue is still there. 

It looks like the page reload is caused by the `TopNav` component in `navigation-component` rendering all menu links as `<a>` elements. Unfortunately I don't think I will be able to investigate this further in the time remaining -- you can close this PR and unassign me.

luizrrodrigues:
@lucyji1000 Okay no problem, thanks for trying this. :)

PR #2657, opened 2019-06-24
Fix sorting in MM leaderboard

ThomasKranitsas:


sushilshinde:
@ThomasKranitsas is this branch`fix-mm-leaderboard-sort` tested separately on any of the deployment enviroment?

ThomasKranitsas:
@sushilshinde if I recall correctly, it was tested on beta

PR #2936, opened 2019-07-18
issue 2837 fix

PrakashDurlabhji:


sumitdaga:
@PrakashDurlabhji i think instead of adding an exception to eslint its better to change the div tag around the search to a button tag or something ...

PrakashDurlabhji:
@sumitdaga PR updated

Issue #2755, opened 2019-07-17
[$15][MSFT-33] [Challenge Feed] User can't select the 'Filters' using tab


veshu:
**Steps to Reproduce**
1. Go to https://www.topcoder.com/challenges and Login using valid credentials
2. Press Tab key till filter is selected

**Expected Result**
User should be able to select the filter by using the tab key

**Actual Result**
User can't select the 'Filters' using tab Key

**Device**: Desktop/Labtop
**Operating System**: Windows
**Browser**: Chrome

**WCAG Category**: 2. Operable
**WCAG Level**: A
**WCAG Success Criterion**: 2.1.1 Keyboard
**Screen Reader Used**: N/A
**Accessibility Audit Tool used**: Other

**Attachment**
https://topcodermsft-my.sharepoint.com/personal/pd-topcoder_topcodermsft_onmicrosoft_com/Documents/Forms/All.aspx?cid=58da9b68%2Df2a3%2D45d2%2Dbbae%2D15abdfd0865b&FolderCTID=0x0120005C598A51049FC14CBA882E1AEE168F51&sortField=Modified&isAscending=true&id=%2Fpersonal%2Fpd%2Dtopcoder%5Ftopcodermsft%5Fonmicrosoft%5Fcom%2FDocuments%2FMSFT%2DTeams%2DQA%2FJune%202019%2F29%20Jun%2FTest%20Results%2FBug%20Videos%20%26%20Log%20Files%2Fcodejam%2DAccessibility%2F18

ThomasKranitsas:
[400]: Failed to create challenge. Detail: Operation failed in the contest service facade.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
[400]: Failed to create challenge. Detail: Operation failed in the contest service facade.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30096156 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

r0hit-gupta:
Please assign me

Oanh-and-only-Oanh:
@r0hit-gupta, please update your tickets with "tcx_readyforreview label if it's ready. thanks

veshu:
I think he is not able to update the ticket till he comes online I will be making those as ready for review.

[...]

PR #2938, opened 2019-07-18
MSFT-33

r0hit-gupta:
changes for #2755


vl-ti:
@r0hit-gupta 
Good work. Your fix works. But there is a minor issue with coding style. Could you please define number constants in curly brackets, not in double-quotes, to keep the style consistent with the existing code. Thanks!

PR #2950, opened 2019-07-18
issue 2886 fix

PrakashDurlabhji:


gurmeetb:
@PrakashDurlabhji same issue

gurmeetb:
@PrakashDurlabhji `Contrast ratio should be 4.5:1` - check using Google Chrome extension. 

2886_Contrast

PrakashDurlabhji:
@gurmeetb i guess this issue is only to those whose topcoder handle name is in grey color, and not for other limits.
ticket mentioned to just change grey color and not others. 

can you please confirm?

PrakashDurlabhji:
@gurmeetb I will update other values too, but can you please provide hex values for each limit, because I installed that extension, but not able to understand how to use it.

gurmeetb:
@PrakashDurlabhji to use it, when on that page click extension, it will show tabular info for Accessibility like it does, each of which matches the respective fields on page. So if that has green check means it is valid based on Accessibility contrast. Hope this helps.

gurmeetb:
@veshu please suggest @PrakashDurlabhji ☝️ on what are the alternative options, if any.

veshu:
@gurmeetb @PrakashDurlabhji For color contrasts issues we can get help from @Dara-K where she will guide to choose the color without loosing branding of the site.

@PrakashDurlabhji Please enroll the @Dara-K if you need the color that can be used.

Dara-K:
Hi @PrakashDurlabhji ,

For the gray handle please use  #555555 and for the green handle #2D7E2D.

codeMinter:
@veshu this looks good, please deploy for QA verification.

PR #2951, opened 2019-07-18
Issue 2885 fix

PrakashDurlabhji:


gurmeetb:
@veshu this looks good, kindly help move to QA for testing.

gurmeetb:
@veshu Sorry for passing this, please hold. I will revert the merge. 

@PrakashDurlabhji please use define a new one in https://github.com/topcoder-platform/topcoder-react-ui-kit and use the variable from there to be constant with how we manage this.

PrakashDurlabhji:
@gurmeetb will it be ok if  i change $tc-dark-blue-70 itself?

this variable may be used at many places so even those color will change? will it be ok to change other places too and not just FORUM POSTS?

Issue #2891, opened 2019-07-17
[$10][MSFT-233] [Challenge Feed][NVD] Hyperlinks are not descriptive to the screen-readers


veshu:
**Steps to Reproduce**
1. Go to https://www.topcoder.com/challenges/30095104
2. Check the 'Here' links

**Expected Result**
- Read about the policy. (Hyperlink the policy text)
- Learn more about how to pass screening. (Hyperlink the pass screening ink)

**Actual Result**
- Read about the policy here.
- Learn more about how to pass screening here.

**Device**: Desktop/Labtop
**Operating System**: Windows
**Browser**: Chrome

**WCAG Category**: 2. Operable
**WCAG Level**: A
**WCAG Success Criterion**: 2.4.4 Link Purpose (In Context)
**Screen Reader Used**: Other
**Accessibility Audit Tool used**: N/A

**Attachment**
https://topcodermsft-my.sharepoint.com/personal/pd-topcoder_topcodermsft_onmicrosoft_com/Documents/Forms/All.aspx?cid=58da9b68%2Df2a3%2D45d2%2Dbbae%2D15abdfd0865b&FolderCTID=0x0120005C598A51049FC14CBA882E1AEE168F51&sortField=Modified&isAscending=true&id=%2Fpersonal%2Fpd%2Dtopcoder%5Ftopcodermsft%5Fonmicrosoft%5Fcom%2FDocuments%2FMSFT%2DTeams%2DQA%2FJune%202019%2F29%20Jun%2FTest%20Results%2FBug%20Videos%20%26%20Log%20Files%2Fcodejam%2DAccessibility%2F69

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30096021 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30096021 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

romitgit:
Please assign this to me.

romitgit:
I'm not able to add tcx_ReadyForReview label

codeMinter:
@romitgit kindly create new PR, your PR had lot of other changes, and main fix for this ticket was not there, we need PR with ticket specific changes only. Not sure why your PR had 15 file changes. 

Please create new PR with fix for this ticket.

[...]

PR #3006, opened 2019-07-24
MSFT-233 Issue#2891

romitgit:
Issue #2891


romitgit:
@codeMinter Here is the new PR which you asked for with the said changes. Please review. Thanks.

codeMinter:
will do in sometime - thanks

Issue #2828, opened 2019-07-17
[$20][MSFT-131] [Challenge Page] Prizes section is not readout by the screen reader


veshu:
**Steps to Reproduce**
1. Go to https://www.topcoder.com/challenges/30095032
2. Switch on the screen reader
3. Tap on the page header and press tab button until it reaches to the prizes section
2. Listen

**Expected Result**
It should be read-out by the screen reader

**Actual Result**
Prizes section is not read out by the screen reader

**Device**: Desktop/Labtop
**Operating System**: Windows
**Browser**: Chrome

**WCAG Category**: 1. Perceivable
**WCAG Level**: A
**WCAG Success Criterion**: 1.3.1 Info and Relationships
**Screen Reader Used**: JAWS
**Accessibility Audit Tool used**: N/A

**Attachment**
https://topcodermsft-my.sharepoint.com/personal/pd-topcoder_topcodermsft_onmicrosoft_com/Documents/Forms/All.aspx?cid=6aa53cde%2D1eef%2D4bae%2Dbfe5%2Df5f5f79e9df4&FolderCTID=0x0120005C598A51049FC14CBA882E1AEE168F51&id=%2Fpersonal%2Fpd%2Dtopcoder%5Ftopcodermsft%5Fonmicrosoft%5Fcom%2FDocuments%2FMSFT%2DTeams%2DQA%2FJune%202019%2F29%20Jun%2FTest%20Results%2FBug%20Videos%20%26%20Log%20Files%2Fcodejam%2DAccessibility%2F41

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30095960 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30095960 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

cwdcwd:
Contest https://www.topcoder.com/challenges/30095960 has been updated - it has been assigned to PkDurlabhji.<br/><br/>```This is an automated message for lazybaer via Topcoder X```

PrakashDurlabhji:
@veshu PR https://github.com/topcoder-platform/community-app/pull/2949

cwdcwd:
[...]

PR #3157, opened 2019-07-29
MSFT-131

r0hit-gupta:
changes for #2828


veshu:
@r0hit-gupta I am not able to use tab key to reach there. tabIndex should be added?
second, after i added tabINdex it said "prize $ 1500, group". There must be contentinfo role for this instead of default group.

r0hit-gupta:
@veshu please check now. Used role `figure` instead of `contentinfo`.

Issue #2823, opened 2019-07-17
[$40]{MSFT-125} Sample challenge page: Screen reader reads the registrants information in a textual format and not in a tabular format.


veshu:
**Steps to Reproduce**
1. Go to the sample challenge page: https://clicktime.symantec.com/34CaebazimgmFxdkBZnihRc7Vc?u=https%3A%2F%2Fwww.topcoder.com%2Fchallenges%2F<sample-Sample challenge page:challengeid>?tab=details. 2. Ensure that the screen reader is switched on. 3. Navigate to the 'registrants' control using the tab key or the arrow key. 4. Press enter on the 'registrants' control. 6. Press the down arrow to access the registrants information. 7. The information is displayed in a tabular form, however screen reader doesn't recognise the tabular format.

**Expected Result**
Native tabular structure should be used so as the screen reader reads the data in a tabular format.

**Actual Result**
Screen reader reads the registrants information in a textual format and not in a tabular format.

**Device**: Desktop/Labtop
**Operating System**: Windows
**Browser**: Chrome

**WCAG Category**: 1. Perceivable
**WCAG Level**: A
**WCAG Success Criterion**: 1.3.1 Info and Relationships
**Screen Reader Used**: JAWS
**Accessibility Audit Tool used**: N/A

**Attachment**
https://topcodermsft-my.sharepoint.com/personal/pd-topcoder_topcodermsft_onmicrosoft_com/Documents/Forms/All.aspx?cid=088f5f5d%2D9b1c%2D4b7e%2Dacbf%2D320c257c9a9a&FolderCTID=0x0120005C598A51049FC14CBA882E1AEE168F51&id=%2Fpersonal%2Fpd%2Dtopcoder%5Ftopcodermsft%5Fonmicrosoft%5Fcom%2FDocuments%2FMSFT%2DTeams%2DQA%2FJune%202019%2F29%20Jun%2FTest%20Results%2FBug%20Videos%20%26%20Log%20Files%2FKameshwarid%2Fbug%2023

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30095999 has been created for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

ThomasKranitsas:
Contest https://www.topcoder.com/challenges/30095999 has been updated - the new changes has been updated for this ticket.<br/><br/>```This is an automated message for thomaskranitsas via Topcoder X```

[...]

PR #3166, opened 2019-07-30
MSFT-125

r0hit-gupta:
Changes for #2823

### DIFF
--- a/src/shared/components/challenge-detail/Registrants/index.jsx
+++ b/src/shared/components/challenge-detail/Registrants/index.jsx
@@ -60,8 +60,8 @@ export default function Registrants({ challenge, checkpointResults, results }) {
 
   return (
     <div styleName={`container ${twoRounds ? 'design' : ''}`} role="table" aria-label="Registrants">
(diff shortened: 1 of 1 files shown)


veshu:
@r0hit-gupta please check below:
1. Table should be aria-label with "Registrants"
2. Design is broken now, please check the live version to compare
3. https://www.w3.org/TR/wai-aria-practices/examples/table/table.html Please see here for more info
4. When I reach to second row it still says 0 rows(for all rows)

r0hit-gupta:
@veshu try this now. Also, please let me know which screen reader are you using. I have used https://www.w3.org/TR/wai-aria-practices/examples/table/table.html as a reference.

veshu:
@r0hit-gupta I am using NVDA and mac(Voice Over) too. I will check the updates. Thanks!

veshu:
@r0hit-gupta Using NVDA it doesn't read the column name of 2 and 3 when reading cell value.

For username it says column "username" but not for other columns it just read as "column 2" and "Column 3" respectively 

https://streamable.com/vrehg

PR #3189, opened 2019-08-01
MSFT-86 | Update tooltip for review scorecard

r0hit-gupta:
Changes for https://github.com/topcoder-platform/community-app/issues/2797


veshu:
@r0hit-gupta This tooltip should be focusable with tab similar to other tooltip above it.

PR #3303, opened 2019-09-09
Mm leaderboard misc fixes v2

ThomasKranitsas:


sushilshinde:
for challenge  30100163

Issue #3520, opened 2019-12-17
$15 | Open new tab for "Payments"


Oanh-and-only-Oanh:
For logged in state:

**Reported bug:** When user clicks on "Payment" in the submenu, they are directed to https://community.topcoder.com/PactsMemberServlet?module=PaymentHistory&full_list=false

**Expected behavior:**
When user clicks on "Payment" in the submenu, a separate tab should open that takes them to https://community.topcoder.com/PactsMemberServlet?module=PaymentHistory&full_list=false

luizrrodrigues:
#3558 
https://github.com/topcoder-platform/navigation-component/pull/82

luizrrodrigues:
@gets0ul Open in new tab works fine, but if back to origin page the Payments menu is active.

So when link will open in new tab we need make sure to not change status of link to active.

![image](https://user-images.githubusercontent.com/7029641/72357432-bdde2480-36c9-11ea-8511-f708bb5c926d.png)

gets0ul:
@luizrrodrigues PR is updated.

nithyaasworld:
Link is not active. However its in Payments menu-item with grey selection indicator.

@luizrrodrigues @Oanh-and-only-Oanh Is this okay?

Here is the screenshot:

Oanh-and-only-Oanh:
Can you please share screenshot?

<https://www.topcoder.com/>

*Oanh Tran*

Product Manager, Community App

732-353-9457

www.topcoder.com  | @topcoder <http://twitter.com/topcoder>

SathyaJayabal:
Verified in beta.

macs054:
Verified in Prod

PR #3558, opened 2020-01-09
Fix for issue #3520 Open New Tab for Payments

gets0ul:
#3520   

Add new openNewTab field into header menu item configuration to support opening a certain menu item in a new tab.


luizrrodrigues:
a4082128a4b9232c481eb05cb8a047be010da639

PR #4249, opened 2020-04-14
Challenge 30121747

simranb86:


simranb86:
@ThomasKranitsas FYI

simranb86:
@ThomasKranitsas I think the builds are failing as it needs to build topcoder-react-lib first and update version in PR, which I have no clue on deployment process.

PR #4385, opened 2020-05-19
Adding markdown view for v5 challenge listings.

rootelement:
Default parses HTML if `descriptionFormat` exists with a value of 'markdown', it will use the react-markdown `Markdown` component to display.


sushilshinde:
https://github.com/topcoder-platform/community-app/issues/4450

Issue #4758, opened 2020-08-15
[$40] View submissions of Completed design contest loads a blank page


SathyaJayabal:
View submissions of Completed design contest loads a blank page
example: 
user: CustomerUser

https://beta-community-app.topcoder.com/challenges/30136817/my-submissions
https://www.topcoder.com/challenges/ae7f8a3f-d28b-4743-ac0f-21e15b9f4891/my-submissions
<img width="1440" alt="Screenshot 2020-08-15 at 3 02 12 PM" src="https://user-images.githubusercontent.com/58783823/90309698-58c10a80-df08-11ea-8a8f-4c7754f3d855.png">

<img width="1440" alt="Screenshot 2020-08-15 at 2 31 53 PM" src="https://user-images.githubusercontent.com/58783823/90309218-4218b480-df04-11ea-9758-7171e11222cc.png">

cc @lakshmiathreya

crazyk07:
Contest https://www.topcoder.com/challenges/30140432 has been created for this ticket.<br/><br/>```This is an automated message for crazyk via Topcoder X```

SathyaJayabal:
@topcoder-platform/topcodercompetitors @topcoder-platform/tcxcommunity this is open for pickup

Submit in 8 hours and PR to

https://github.com/topcoder-platform/community-app/tree/milestone-v1.3
https://github.com/topcoder-platform/topcoder-react-lib/tree/milestone-v1.3

run npm run test before PR

crazyk07:
Contest https://www.topcoder.com/challenges/30140432 has been updated - it has been assigned to cagdas001.<br/><br/>```This is an automated message for crazyk via Topcoder X```

cagdas001:
PR: #4857

luizrrodrigues:
@cagdas001 My Submissions page missing 2 information:
- **The challenge has ended** - instead that showing **0H 0M left**
- **Ends: Thursday 09/03/20 10:06 PM**

![image](https://user-images.githubusercontent.com/7029641/92188121-18cbb400-ee32-11ea-8645-059c981f9a9f.png)

sandhiyakavi:
Tested.View Submission loads the my submission page successfully.

![image](https://user-images.githubusercontent.com/55895538/92204635-e1efaf80-eebe-11ea-9086-44b93b6462a6.png)
![image](https://user-images.githubusercontent.com/55895538/92204747-1a8f8900-eebf-11ea-9dbe-368c76311b88.png)

[...]

PR #4857, opened 2020-09-03
fix(submission-management): add null checking for `currentPhase`

cagdas001:
Add null checking when rendering properties from `currentPhase`

Reference #4758

### DIFF
--- a/src/shared/components/SubmissionManagement/SubmissionManagement/index.jsx
+++ b/src/shared/components/SubmissionManagement/SubmissionManagement/index.jsx
@@ -86,7 +86,7 @@ export default function SubmissionManagement(props) {
             )
           }
           {
-            challenge.status !== 'COMPLETED' ? (
+            challenge.status !== 'Completed' ? (
               <div>
                 <p styleName="time-left">
                   {days > 0 && (`${days}D`)}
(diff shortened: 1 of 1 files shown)


src/shared/components/SubmissionManagement/SubmissionManagement/index.jsx
```diff
@@ -78,9 +78,13 @@ export default function SubmissionManagement(props) {
...
-          </p>
+          {
+            currentPhase && (
+            <p styleName="round">
+              {currentPhase.name}
+            </p>
+            )
+          }
           {
             challenge.status !== 'COMPLETED' ? (
```
luizrrodrigues, line 89: Check this validation, this was changed from `COMPLETED` to `Completed`

luizrrodrigues (changes requested):
Please check feedback here: https://github.com/topcoder-platform/community-app/issues/4758#issuecomment-686841502

src/shared/components/SubmissionManagement/SubmissionManagement/index.jsx
cagdas001: > The challenge has ended - instead that showing 0H 0M left

Looks like its because of this line, changing this to `Completed` should solve it

> Ends: Thursday 09/03/20 10:06 PM

After my changes, if challenge is ended, you should not see this text. And if it's visible (means challenge still open), it should show phase name as well: `<PhaseName> Ends Thursday 09/03/20 10:06 PM` So, you'll either see nothing at there, or will see the phase name as well. Where did you get this?

luizrrodrigues: @cagdas001 Yep, looks right

cagdas001: @luizrrodrigues pushed the new commit for `COMPLETED` -> `Completed` change, PR will be available after automated tests

luizrrodrigues (approved):

PR #4861, opened 2020-09-04
Implement Generate Document Links

fikzzzy:


luizrrodrigues (changes requested):
This solution not valid, this getting links from challenge description.

Download items not related with description, need investigate why `documents` are missing from challenge data.

Check live site before API v5 integration:
https://beta-community-app.topcoder.com/challenges/30129332

fikzzzy:
> This solution not valid, this getting links from challenge description.
> 
> Download items not related with description, need investigate why `documents` are missing from challenge data.
> 
> Check live site before API v5 integration:
> https://beta-community-app.topcoder.com/challenges/30129332

@luizrrodrigues so we need to modify the backend api in topcoder-platform/challenge-api ?

luizrrodrigues:
@fikzzzy If the issue is with API, please add the label  to issue

Issue #4294, opened 2020-04-22
[$40] UI bug with dates window on profile settings page


Oanh-and-only-Oanh:
Bug description: 
1) go to https://www.topcoder.com/settings/profile/#work
2) When trying to pick the dates of my job, the date-choice window is falling below and behind the bottom of the page in my Chrome, and does not allow to make the choice

Expected behavior:
1) go to https://www.topcoder.com/settings/profile/#work
2) When trying to pick the dates of my job, the date-choice window should be fully visible to make the date choice

SathyaJayabal:
<img width="1440" alt="Screenshot 2020-04-28 at 2 43 21 PM" src="https://user-images.githubusercontent.com/58783823/80469824-bed85080-895e-11ea-97ca-bfc99ad1d4bc.png">

Jobee1:
The issue persists for Skill dropdown too.
![Screenshot_31](https://user-images.githubusercontent.com/48575864/80477451-a7ec2b00-896b-11ea-8f5b-32c9356a8759.png)

crazyk07:
Contest https://www.topcoder.com/challenges/30141152 has been created for this ticket.<br/><br/>```This is an automated message for crazyk via Topcoder X```

luizrrodrigues:
@topcoder-platform/topcodercompetitors @topcoder-platform/tcxcommunity this is open for pickup

Submit in 8 hours and PR to

https://github.com/topcoder-platform/community-app/tree/milestone-20200924
https://github.com/topcoder-platform/topcoder-react-lib/tree/milestone-20200924

run `npm run test` before PR

crazyk07:
Contest https://www.topcoder.com/challenges/30141152 has been updated - it has been assigned to gets0ul.<br/><br/>```This is an automated message for crazyk via Topcoder X```

luizrrodrigues:
@gets0ul Please check feedback here: https://github.com/topcoder-platform/community-app/pull/4892#pullrequestreview-487351673

gets0ul:
@luizrrodrigues Fixed.

sandhiyakavi:
Verified on staging(Develop).Working as Expected

![image](https://user-images.githubusercontent.com/55895538/93422021-d30cf200-f8ed-11ea-8ca6-22525484e5e5.png)
![image](https://user-images.githubusercontent.com/55895538/93422038-e029e100-f8ed-11ea-9eeb-80f8e57e08b4.png)
[...]

PR #4892, opened 2020-09-10
Fix for Issue #4294 Dropdown hidden behind on profile settings page

gets0ul:
#4294: Fix dropdown issue being hidden behind footer on profile settings page.

### DIFF
--- a/src/shared/components/Header/index.jsx
+++ b/src/shared/components/Header/index.jsx
@@ -5,6 +5,8 @@ import { config } from 'topcoder-react-utils';
 import Logo from 'assets/images/tc-logo.svg';
 import { tracking } from '../../actions';
 
+import './style.scss';
+
 let TopNavRef;
 let LoginNavRef;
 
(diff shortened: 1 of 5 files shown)


src/shared/components/Header/index.jsx
```diff
@@ -69,7 +69,7 @@ const Header = ({
 
   if (TopNavRef) {
     return (
-      <div>
+      <div style={{ zIndex: 2 }}>
```
luizrrodrigues, line 72: @gets0ul Please don't use inline styles, move this to .scss file.

luizrrodrigues (changes requested):
Styles need be added to .scss files instead .jsx

src/shared/components/Header/index.jsx
gets0ul: @luizrrodrigues Done.

luizrrodrigues (approved):

PR #5119, opened 2020-10-16
Fix Empty Profile Picture

fikzzzy:
image


luizrrodrigues (changes requested):
@fikzzzy The issue is about `tourist` profile showing placeholder image instead photo.

image

PR #5254, opened 2020-12-10
Filter update -- fix issue 5068's comments

nqviet:
**issuecomment-741596441**:  _(1) intentional_, _(2) fixed_: layout issue, _(3) fixed_: after coming back from the challenge details, the default tab is 'Active' and default bucket is 'Open for registration'
**issuecomment-741600474**: _fixed_: layout issue
**issuecomment-741904592**: _fixed_: forgot to set the challenge 'types' filter to all when clicking on the challenge Tags

&nbsp;

About 741904592:  `search=<tag>` vs `tags[]=<tag>`
- There are modules of the community-app allowing to search by multiple tags (ex: Contentful EDU search) while `search=<tag>`  is allowing searching by a single tag only
- `search=<tag>` searching in challenges' name, description and tags, so there is a confused case that a challenge with only HTML tag but its description including the keyword 'Java' still being listed when user clicking on the 'Java' tag under the header of some challenge details.

So I think `tags[]=<tag>` is better than `search=<tag>` and its search results will be the same as the results of the old filter panel.

### DIFF
--- a/__tests__/shared/components/challenge-listing/__snapshots__/index.jsx.snap
+++ b/__tests__/shared/components/challenge-listing/__snapshots__/index.jsx.snap
@@ -24,18 +24,6 @@ exports[`Matches shallow shapshot 1 shapshot 1 1`] = `
         setFilterState={[MockFunction]}
       />
     </div>
-    <div
-      className="src-shared-components-challenge-listing-___style__sidebar-container-desktop___h0bz6"
-    >
-      <Connect(SidebarContainer)
-        expanding={false}
-      />
-      <Connect(Container)
-        communityName={null}
-        isAuth={false}
-        setFilterState={[MockFunction]}
-      />
-    </div>
     <Connect(Listing)
       activeBucket="abc"
       auth={Object {}}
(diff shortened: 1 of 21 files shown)


maxceem:
> There are modules of the community-app allowing to search by multiple tags (ex: Contentful EDU search) while search=<tag> is allowing searching by a single tag only

@nqviet thank you for suggestions regarding tags. Could you please, share some link/screenshot where we have a search by multiple tags, cannot find such an example by myself.

> search=<tag> searching in challenges' name, description and tags, so there is a confused case that a challenge with only HTML tag but its description including the keyword 'Java' still being listed when user clicking on the 'Java' tag under the header of some challenge details.

This is an intentional change. We are making search more "intelligent" type into search field what you need, and we would show it. This is to get rid of the complex search UI.

maxceem (changes requested):
Fixes look good. The only thing there is no way to close the calendar in mobile view anymore. Before there was a "close" icon in the top right corner.

image

nqviet:
> > There are modules of the community-app allowing to search by multiple tags (ex: Contentful EDU search) while search= is allowing searching by a single tag only
> 
> @nqviet thank you for suggestions regarding tags. Could you please, share some link/screenshot where we have a search by multiple tags, cannot find such an example by myself.

@maxceem  There is still one place where `/?tags[]=[<tag1>]&tags[]=<tag2>` is still applied in the Contentful module. 


Untitled

maxceem (approved):
Thanks for the quick update, @nqviet

It looks like the place you've mentioned in the code is not related to challenges searching but is for searching articles in Thrive, see screenshot https://monosnap.com/file/fHAMyTfc2cvzvBSYlje5bHRJfKKwog.

So we should be safe to use "search" for filtering challenges by tags.

So I'm passing this to QA team, to check if all is good now.

Issue #5599, opened 2021-06-28
New Navi component for TCO22


kkartunov:
Desktop: https://marvelapp.com/prototype/ccah4h1/screen/80447499
Mobile: https://marvelapp.com/prototype/666a7di/screen/80484873

kkartunov:
@SathyaJayabal this one can be tested on https://beta-community-app.topcoder.com/examples/contentful/viewport/1nvm6EsPMHPLBtuGXNRMYV/

- ignore that it loads the same page all the time, we focus on the top header Navi itself in this issue

SathyaJayabal:
@kkartunov for mobile the current sub menu selection is not highlighted

for example, this is the marathon page under competition, menu, but the marathon nav item is not highlighted.
![WhatsApp Image 2021-07-07 at 4 14 32 PM](https://user-images.githubusercontent.com/58783823/124746432-79d68380-df3e-11eb-973a-2749d67ddf81.jpeg)

kkartunov:
@SathyaJayabal it is as per design but the team is looking into that. Please hold ...

sdgun:
@kkartunov Verified in mobile. Current sub menu selection is highlighted now. Also checked in desktop.

![image](https://user-images.githubusercontent.com/4592712/125029881-e91bb700-e0a7-11eb-8e5e-58bf5c859074.png)

sandhiyakavi:
@kkartunov If we click the user handle dropdown at the top header navi , the pop up menu is getting hided behind the image.

![image](https://user-images.githubusercontent.com/55895538/125231220-4c5e5100-e315-11eb-9d3b-601970577f8b.png)

Also under Regionals menu, only the submenu Overview is getting displayed, but in the design I am able to see many sub menus. is it expected?

![image](https://user-images.githubusercontent.com/55895538/125231462-becf3100-e315-11eb-88e9-7063b8bd1fd6.png)

![image](https://user-images.githubusercontent.com/55895538/125231439-b24ad880-e315-11eb-86a1-b9834db14fa2.png)

kkartunov:
@sandhiyakavi this issues with z-index is fixed.

sandhiyakavi:
Verified on Staging Env. Issue resolved and the Header navi in both desktop as well as Mobile is good.

![image](https://user-images.githubusercontent.com/55895538/125261749-8858dc80-e33c-11eb-8971-e8bd6eea6433.png)

sandhiyakavi:
[...]

PR #5637, opened 2021-07-09
Navi tco

kkartunov:
#5599


package.json
```diff
@@ -88,7 +88,8 @@
     "money": "^0.2.0",
     "morgan": "^1.9.0",
     "multer": "^1.4.2",
-    "navigation-component": "topcoder-platform/navigation-component#develop",
+    "navigation-component": "github:topcoder-platform/navigation-component#develop",
+    "navigation-component-tco": "github:topcoder-platform/navigation-component-tco#new-dev",
```
luizrrodrigues, line 92: @kkartunov We going to use this component with `new-dev`? Maybe we can create a release from `navigation-component-tco` and use that instead github link?

luizrrodrigues (approved):

Issue #5781, opened 2021-11-12
[$10] Even though first name last name are required, my profile page shows empty fields


LieutenantRoger:
**Summary** 
Even though first name last name are required, my profile page shows empty fields . To be clear , I have entered my first name , last name on my TC Profile

**Issue Type**
Functional

**Steps**
1. Open https://www.topcoder.com/settings/profile
2. Observe on my profile there is no firstname, lastname even though it shows that they are required

**Actual result**: Even though first name last name are required, my profile page shows empty fields 

**Expected results**: Should not allow no firstname,lastname

**Snapshots/Recording**
![is1](/uploads/873f52e8a64afc22ab9d4c9f56363874/is1.png)

***Device/Broswer**: 
Browser: Chrome https://prnt.sc/1vb5r8v
Device: Windows 10 https://prnt.sc/1vb5v0a

LieutenantRoger:
Challenge https://www.topcoder.com/challenges/7cf3a9f3-6b0e-450a-93b2-507b22d8c98e has been created for this ticket.<br/><br/>```This is an automated message for lieutenantroger via Topcoder X```

shivam-51:
@bug-bash-helper assign

shivam-51:
@bug-bash-helper assign

LieutenantRoger:
Challenge https://www.topcoder.com/challenges/7cf3a9f3-6b0e-450a-93b2-507b22d8c98e PkDurlabhji has been unassigned.<br/><br/>```This is an automated message for lieutenantroger via Topcoder X```

LieutenantRoger:
Challenge https://www.topcoder.com/challenges/7cf3a9f3-6b0e-450a-93b2-507b22d8c98e has been assigned to shivam51.<br/><br/>```This is an automated message for lieutenantroger via Topcoder X```

shivam-51:
@bug-bash-helper unassign

LieutenantRoger:
Challenge https://www.topcoder.com/challenges/7cf3a9f3-6b0e-450a-93b2-507b22d8c98e shivam51 has been unassigned.<br/><br/>```This is an automated message for lieutenantroger via Topcoder X```

shivam-51:
@bug-bash-helper assign

LieutenantRoger:
[...]

PR #5824, opened 2021-11-19
fix: #5781

shivam-51:


shivam-51:
My solution: Show Error message for empty FirstName and LastName even on the existing empty names.

Issue #5795, opened 2021-11-12
[$10] Can upload any type of file to Profile picture, no file validation


LieutenantRoger:
**Summary** 
Can upload any type of file to Profile picture

**Issue Type**
Functional

**Steps**
1. Open https://www.topcoder.com/settings/profile
2. Click upload a new avatar and select All Files from File Type and select a .zip file

**Actual result**: Can upload any type of file to Profile picture, no file validation

**Expected results**: Should have file validation

**Snapshots/Recording**
![Topcoder_-_Google_Chrome_2021-10-22_20-40-24_Trim](/uploads/343e470843d2b9e68ac3fa2d338df8b5/Topcoder_-_Google_Chrome_2021-10-22_20-40-24_Trim.mp4)

***Device/Broswer**: 
Browser: Chrome https://prnt.sc/1vb5r8v
Device: Windows 10 https://prnt.sc/1vb5v0a

LieutenantRoger:
Let's restrict the frontend can only upload normal pictures such as: 'jpg, png, jpeg' files

LieutenantRoger:
Challenge https://www.topcoder.com/challenges/bdb20225-cb77-4665-b0b3-2c9a8d8cbb9a has been created for this ticket.<br/><br/>```This is an automated message for lieutenantroger via Topcoder X```

FilipRazek:
@bug-bash-helper assign

FilipRazek:
@bug-bash-helper [{link to pull request}](https://github.com/topcoder-platform/community-app/pull/5829) is ready for review

LieutenantRoger:
https://user-images.githubusercontent.com/2875203/142762132-53e6af10-ee69-4f42-b282-c476cbef9e0e.mp4

1. The fix doesn't work, please see the recording on my local. 
2. Please create your working branch based on `ca-profile-bug-bash` branch,  you seems creating the branch based on 'develop' branch, also your PR is targeting the `develop` branch , which are not correct

shivam-51:
@bug-bash-helper assign

shivam-51:
Is this issue not open for pickup anymore?

LieutenantRoger:
Resigning @FilipRazek as theres' not progress.  

@shivam-51 I will assign you, and please let me know if you will work on this. :)

LieutenantRoger:
@shivam-51 In case I missed to respond in the ticket, can message me via slack. It's not easy for me to see the comments among so many emails. 

[...]

PR #5829, opened 2021-11-20
[Bug Bash] Frontend image format validation (#5795)

FilipRazek:
Allow only images to be uploaded as avatar


LieutenantRoger:
This seems a invalid PR, closing it

PR #5910, opened 2021-12-06
Contentful app component docs

kkartunov:
Update Contentful Docs


kkartunov:
@luizrrodrigues merging this to dev as it is only Contentful docs update it does not require any QA/realease attention. Let's just let it propagate upstream...

PR #6334, opened 2022-05-10
additional fixes from challenge listing reskin

nursoltan-s:


dedywahyudi:
@nursoltan-s There's test failed. Can you check please?

PR #6338, opened 2022-05-11
Implement winner tab

52cs:
close Response 1: Comment ([lunarkid](http://www.topcoder.com/tc?module=MemberProfile&cr=22706873))
- Avatar need pulled from the api, refer the usage to Profile page
- User with no profile image not correct, follow Figma Design with background

This is my rejected submission.
It seems that it satisfies the requirements after the requirements change.
https://discussions.topcoder.com/discussion/comment/28847


52cs:
change PR branch in https://github.com/topcoder-platform/community-app/pull/6339

PR #6339, opened 2022-05-12
Implement winner tab

52cs:
Re-PR: fix PR-Branch

> close Response 1: Comment ([lunarkid](http://www.topcoder.com/tc?module=MemberProfile&cr=22706873))
> 
> * Avatar need pulled from the api, refer the usage to Profile page
> * User with no profile image not correct, follow Figma Design with background
> 
> This is my rejected submission. It seems that it satisfies the requirements after the requirements change. https://discussions.topcoder.com/discussion/comment/28847


dedywahyudi:
@52cs 

PR only for challenge winner. I need close this. 

Your previous version still not correct apart from the profile image. 
You not use correct placeholder of user image based on Figma Design. 

Check feedback on challenge.

dedywahyudi:
@52cs 
You should NOT create PR. 
PR only for the winner. 
Create patch & submit to challenge

PR #6381, opened 2022-06-09
enhance mm dashboard

nursoltan-s:


rakibansary (changes requested):
Hi @nursoltan-s,

The update looks good. One small issue - sometimes the highlighted points are overlapped by other points

image

Can we try an implementation where each submitters points are added as an array. And on click of the selected point, the group is brought to front. Please see the following fiddle for an example of this.

https://jsfiddle.net/krt93ns6/

PR #6750, opened 2022-12-08
updated sprig for dev testing using events

jpeg22:
Makes sprig event call when handleSubmit is run on submission page.

### DIFF
--- a/src/shared/containers/SubmissionPage.jsx
+++ b/src/shared/containers/SubmissionPage.jsx
@@ -70,7 +70,7 @@ class SubmissionsPageContainer extends React.Component {
       track,
     } = this.props;
     // When the user is waiting for their submission to upload, the survey should appear
-    Sprig('track', 'onUploadSubmission'); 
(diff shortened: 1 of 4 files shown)


src/shared/containers/SubmissionPage.jsx
```diff
@@ -64,7 +69,8 @@ class SubmissionsPageContainer extends React.Component {
       challenge,
       track,
     } = this.props;
-
+    // When the user is waiting for their submission to upload, the survey should appear
+    Sprig('track', 'onUploadSubmission');
```
brooketopcoder, line 73: If you want to use the sdk, you can do `sprig.track('onUploadSubmission')`.

You might also consider adding the user info to the events.

https://github.com/topcoder-platform/platform-ui/blob/TCA-790_assessments/src-ts/lib/functions/survey-functions/survey.functions.ts

brooketopcoder (approved):
Looks good

PR #6766, opened 2023-01-05
Prod 3521 sprig challenge listings

jpeg22:
Added sprig event to challenge listings page render.


brooketopcoder (approved):
Looks good to me.

PR #6865, opened 2023-04-27
P0 fix for leaking M2M

jmgasper:


src/shared/reducers/contentful/space.js
```diff
@@ -51,7 +51,7 @@ function create(init) {
     const { data, target } = action.payload;
     const res = { ...state, [target]: content(state[target], action) };
 
-    if (action.type === actions.contentful.queryContentDone.toString() && data.includes) {
+    if (action.type === actions.contentful.queryContentDone.toString() && data && data.includes) {
```
testflyjets, line 54: Can we write this as:
```
if (action.type === actions.contentful.queryContentDone.toString() && data?.includes) {
```

src/shared/routes/index.jsx
```diff
@@ -117,7 +117,12 @@ function Routes({ communityId }) {
         <Redirect
           exact
           from="/policy"
-          to="/privacy"
+          to={config.POLICY_PAGES_PATH}
```
testflyjets, line 120: Does `POLICY_PAGES_PATH` already exist as an config setting and/or env var?

PR #6892, opened 2023-06-17
PROD -- fix crushing when contentful items empty, also fix saving profiles without country info (PROD-4287)

eisbilir:
https://topcoder.atlassian.net/browse/PROD-4301
https://topcoder.atlassian.net/browse/PROD-4287


config/qa.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  SEGMENT_IO_API_KEY: 'QBtLgV8vCiuRX1lDikbMjcoe9aCHkF6n',
+  SERVER_API_KEY: '79b2d5eb-c1fd-42c4-9391-6b2c9780d591',
```
luizrrodrigues, line 3: @jmgasper Expose these KEY's is expected?

luizrrodrigues (changes requested):
Expose API KEY's

config/qa.js
jmgasper: @luizrrodrigues - Probably not.  @rakibansary - Can you take a look please?  I think you added these.

rakibansary: @luizrrodrigues @jmgasper, we've always had these

image

But I agree, we could improve our implementation to reduce the risk of exposing such sensitive information. But it might not be a quick lift.

luizrrodrigues: Okay, let follow up with that for now because this already present in dev config.

Can we raise a ticket in Jira to track this?

luizrrodrigues (approved):

PR #6950, opened 2023-12-09
IC-15

jmgasper:
Add new `Innovation Challenges` tab.  When clicked, it will filter to only those challenges that have tag `Innovation Challenge`


kkartunov (approved):
Looks good 👍

PR #6955, opened 2024-01-11
PROD - Fixes to use v5 for challenge results instead of v2

jmgasper:
https://topcoder.atlassian.net/browse/PS-187

### DIFF
--- a/src/shared/containers/challenge-detail/index.jsx
+++ b/src/shared/containers/challenge-detail/index.jsx
@@ -775,7 +775,6 @@ ChallengeDetailPageContainer.propTypes = {
   fetchChallengeStatistics: PT.func.isRequired,
   getAllCountries: PT.func.isRequired,
   getReviewTypes: PT.func.isRequired,
-  // loadResults: PT.func.isRequired,
   // loadingCheckpointResults: PT.bool,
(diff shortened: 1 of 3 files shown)


src/shared/containers/challenge-detail/index.jsx
```diff
@@ -970,8 +969,8 @@ const mapDispatchToProps = (dispatch) => {
             } else dispatch(a.dropCheckpoints());
           } else dispatch(a.dropCheckpoints());
           if (ch.status === CHALLENGE_STATUS.COMPLETED) {
-            dispatch(a.loadResultsInit(ch.legacyId));
-            dispatch(a.loadResultsDone(tokens, ch.legacyId, ch.track.toLowerCase()));
+            dispatch(a.loadResultsInit(ch.id));
+            dispatch(a.loadResultsDone(ch.id));
```
kkartunov, line 973: @jmgasper why are the other arguments `loadResultsDone` missing?
Is `loadResultsDone` moved to the react-lib and updated thus not to see it here?

jmgasper: The tokens weren't used in the original call, and the track isn't necessary since we can now get that from the v5 challenge object response.

kkartunov (approved):

PR #6957, opened 2024-01-17
CORE-201 Shut down Community App - Beta,Staging,Test & QA envs

kkartunov:
https://topcoder.atlassian.net/browse/CORE-201
https://topcoder.atlassian.net/browse/CORE-202
https://topcoder.atlassian.net/browse/CORE-250
https://topcoder.atlassian.net/browse/CORE-251


jmgasper (approved):
Looks good.  I don't see anything that will break staging, but just wanted to make sure this won't affect staging, since we still use that

kkartunov:
@jmgasper actually we are extending this for staging as well.

jmgasper:
> @jmgasper actually we are extending this for staging as well.

We use staging for QA validation of prod fixes.  We can work around not having it, but it is useful at times.

kkartunov:
@testflyjets FYI.

.circleci/config.yml
testflyjets, line 1: Why not just remove this folder and file entirely? We have version control and can bring it back.

testflyjets (approved):
Please add a comment at the top of the README saying this code has been DEPRECATED as of today.

.circleci/config.yml
kkartunov: We need it for prod & dev workflows, those remain.

PR #6976, opened 2024-04-24
PROD - deploy PS-257, conditional links for Wipro users in the old header

jmgasper:
https://topcoder.atlassian.net/browse/PS-257


testflyjets (approved):
Nicely done!

PR #7022, opened 2024-12-18
PM-199, PM-209 Denial of service fix

hentrymartin:
## What's in this PR?

- Used promised based fs functions for getting file content
- Also checking for file size and checking if the path exists

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -83,7 +83,7 @@ const swScope = '/challenges'; // we are currently only interested in improving 
 const tcoPattern = new RegExp(/^tco\d{2}\.topcoder(?:-dev)?\.com$/i);
 const universalNavUrl = config.UNIVERSAL_NAV_URL;
 
-const getExtraScripts = (ts) => ([
+const getExtraScripts = ts => [
   `<script type="application/javascript">
   if('serviceWorker' in navigator){
     navigator.serviceWorker.register('${swScope}/${sw}', {scope: '${swScope}'}).then(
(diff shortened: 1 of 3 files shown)


src/server/index.js
```diff
@@ -42,17 +42,48 @@ global.atob = atob;
...
-ts = moment(ts.timestamp).valueOf();
+const getTimestamp = async () => {
+  let timestamp;
+  try {
+    const filePath = path.resolve(__dirname, '../../.build-info');
+    if (!filePath.startsWith(path.resolve(__dirname, '../../'))) {
+      throw new Error('Invalid file path detected');
+    }
+
+    const MAX_FILE_SIZE = 10 * 1024; // 10 KB max file size
```
jmgasper, line 53: Can we put this out to an env variable and default it in `config/default.js` please?

src/server/services/communities.js
```diff
@@ -30,24 +30,57 @@ async function getGroupsService() {
...
+    const validationPromises = ids.map(async (id) => {
+      const uri = path.resolve(METADATA_PATH, id, 'metadata.json');
+
+      try {
+        // Check if the file exists
+        await fs.promises.access(uri);
+
+        // Get file stats
+        const stats = await fs.promises.stat(uri);
+        const MAX_FILE_SIZE = 1 * 1024 * 1024; // 1 MB
```
jmgasper, line 48: Same here - can we make this configurable with a default in `config/default.js` please?

jmgasper (changes requested):

hentrymartin:
@jmgasper Can you please approve this PR, if there are no more comments to be resolved?

kkartunov (approved):

src/server/index.js
jmgasper: No, not right now - it's fine to leave this as is.

jmgasper (approved):

PR #7026, opened 2025-01-02
fix(PM-208): removed my account screen

hentrymartin:
## What's in this PR?

- Removed account screen and as per Justin this screen is not used any more so removing this would resolve empty password security issue


jmgasper (approved):
Looks good, thanks!

kkartunov (approved):

PR #7027, opened 2025-01-06
fix(PM-204): open redirect issues

hentrymartin:
## What's in this PR?

- Fixed open redirect issues
- Explicitly checking for base urls whether they are valid
- Using URL constructor to construct the url


src/server/services/contentful.js
```diff
@@ -21,6 +21,7 @@ const PREVIEW_URL = 'https://preview.contentful.com/spaces';
 export const ASSETS_DOMAIN = 'assets.ctfassets.net';
 export const IMAGES_DOMAIN = 'images.ctfassets.net';
 
+export const ALLOWED_DOMAINS = ['assets.ctfassets.net', 'images.ctfassets.net'];
```
jmgasper, line 24: Can we make this const use `ASSETS_DOMAIN` and `IMAGES_DOMAIN` from above instead of re-hardcoding the values please/

src/shared/components/TopcoderHeader/index.jsx
```diff
@@ -44,6 +44,7 @@ import './style.scss';
 /* global window, document */
 
 const BASE_URL = config.URL.BASE;
+const VALID_BASE_URLS = ['https://www.topcoder-dev.com', 'https://www.topcoder.com'];
```
jmgasper, line 47: Do we need this?  Can we just base this off of `config.URL.BASE`?

jmgasper (changes requested):
A couple minor tweaks requested, thanks!

src/server/services/contentful.js
jmgasper: Thanks

hentrymartin:
@jmgasper Can you please approve this PR, if there are no more comments to be resolved?

kkartunov (approved):

jmgasper (approved):

PR #7031, opened 2025-01-08
fix(PM-195): Cross site request forgery

hentrymartin:
## What's in this PR?

- Removed some of the api's which are not being used anymore
- This will remove the cross site request forgery vulnerabilities in them


jmgasper (commented):
I was just chatting with a Marathon Match copilot today who mentioned they are planning a year long MM with leaderboards.  Will that still work with the mmLeaderboard service going away?  I'm not quite sure if that service is still used or not...

hentrymartin:
> I was just chatting with a Marathon Match copilot today who mentioned they are planning a year long MM with leaderboards. Will that still work with the mmLeaderboard service going away? I'm not quite sure if that service is still used or not...

@jmgasper I've added back the mml service

jmgasper (approved):

kkartunov (approved):

PR #7033, opened 2025-01-13
PM-201: removed automated smoke test Dockerfile and removed the option from circle ci

hentrymartin:
## What's in this PR?

- Removed automated smoke test dockerfile as it is outdated(Asked Sathya about this and got a reply that it is not used now)
- Also, removed the option from circle ci


kkartunov (approved):
Looks good.

PR #7034, opened 2025-01-14
PM-210: Fix privacy violations

hentrymartin:
## What's in this PR?

- Added checks to check safe tags when rendering contentful pages
- Instead of spreading whole props object in the reported files, changed to only pass the props that are needed inside the child component


kkartunov (approved):
Looks good.

PR #7037, opened 2025-01-20
fix(PM-594, PM-596, PM-606): Regression fixes

hentrymartin:
## What's in this PR?

- Added back thrive endpoint as it is needed for searching thrive articles
- Changed the logic to check for invalid id in recruit CRM service
- Removed styles from the unsafe tags


jmgasper (approved):
Looks good

kkartunov (approved):

PR #7050, opened 2025-01-27
[PROD] Security fixes - release

kkartunov:
As per https://topcoder.atlassian.net/browse/PM-189


Dockerfile
```diff
@@ -4,10 +4,14 @@
 
 FROM node:8.11.2
```
kkartunov, line 5: Should this be `10.24.1` as in the package.json?

PR #7069, opened 2025-02-17
fix(PM-714): encode uri return url to fix xss dom error

hentrymartin:
## What's in this PR?

- Added a check to encode uri component for return url to avoid xss dom attack

### DIFF
--- a/src/shared/containers/tc-communities/tco20/Header.jsx
+++ b/src/shared/containers/tc-communities/tco20/Header.jsx
@@ -25,8 +25,8 @@ function TCO20Header(props) {
           </React.Fragment>
         ) : (
           <React.Fragment>
-            <a href={`${config.URL.AUTH}/member?utm_source=TCO20site&retUrl=${getCurrentUrl()}`} className={defaultStyle.loginLink}>LOGIN</a>
-            <a href={`${config.URL.AUTH}/member/registration?utm_source=TCO20site&retUrl=${getCurrentUrl()}`} className={defaultStyle.signUpLink}>SIGN UP</a>
+            <a href={`${config.URL.AUTH}/member?utm_source=TCO20site&retUrl=${encodeURIComponent(getCurrentUrl())}`} className={defaultStyle.loginLink}>LOGIN</a>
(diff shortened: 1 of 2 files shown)


jmgasper (approved):

kkartunov (approved):

src/shared/containers/tc-communities/tco20/Header.jsx
```diff
@@ -25,8 +25,8 @@ function TCO20Header(props) {
           </React.Fragment>
         ) : (
           <React.Fragment>
-            <a href={`${config.URL.AUTH}/member?utm_source=TCO20site&retUrl=${getCurrentUrl()}`} className={defaultStyle.loginLink}>LOGIN</a>
-            <a href={`${config.URL.AUTH}/member/registration?utm_source=TCO20site&retUrl=${getCurrentUrl()}`} className={defaultStyle.signUpLink}>SIGN UP</a>
+            <a href={`${config.URL.AUTH}/member?utm_source=TCO20site&retUrl=${encodeURIComponent(getCurrentUrl())}`} className={defaultStyle.loginLink}>LOGIN</a>
```
vas3a, line 28: not sure about this, if you change how the "retUrl" is sent, you have to update the code that handles it too. (eg. the receiver should "decodeURIComponent" the retUrl).

Also, not sure if this is considered xss prevention - when we do decodeURIComponent, we still can be vulnerable to XSS. We have to strip out the "vulnerable" part - eg. the query & the hash part. You could use location.origin & location.pathname instead of location.href (eg `${window.location.origin}${window.location.pathname}`).

Just my 2 cents. but if you've used this previously and it fixed the security issue & is safe on the handler side, you can leave it as is 👍

hentrymartin: @vas3a Thanks for the suggestion. I've updated the logic in getCurrentUrl function to frame url using origin and pathname but will also retain the encoding part, I tested this out in TCO20 homepage and the login works fine with the return url being encoded and I think auth0 decodes the url in their side(as it is fairly a good practice to do so).

PR #7071, opened 2025-02-17
fix(PM-716): used promisify instead of sync file read API

hentrymartin:
## What's in this PR?

- Used promisify instead of sync file read API which should resolve path manipulation issue.

### DIFF
--- a/src/server/services/communities.js
+++ b/src/server/services/communities.js
@@ -132,7 +132,7 @@ export async function getMetadata(communityId) {
     communityId, 'metadata.json',
   );
   try {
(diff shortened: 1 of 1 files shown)


jmgasper (approved):

kkartunov (approved):

kkartunov:
@hentrymartin we need to keep eye on this and avoid errors like "cannot read property 'stat' of undefined".
See: https://github.com/topcoder-platform/community-app/commit/542b81eb11a56593f1e2b07e26dd89670afc5d2e
Not sure if related to promisify or fs module but we had this issue in production when released v1 security fixes.

PR #7073, opened 2025-02-19
fix(PM-717): use crypto random instead of math random

hentrymartin:
## What's in this PR?

- Used crypto random instead of math random as it is not secured.


kkartunov (approved):
Looks good.

PR #7091, opened 2025-05-07
fix(PM-1192): showing wrong review scorecard link

hentrymartin:
## What's in this PR?

- In current production version, we are showing the score card based on the legacy object
- The correct score card numbers are coming inside corresponding phases
- This PR changes the logic to use phases instead of legacy

Ticket link - https://topcoder.atlassian.net/browse/PM-1192


kkartunov (approved):
Looks good.

